PR #48, opened 2018-05-22
WIP Custom layouts, campaign templates

idengager:
**Please don’t merge yet.**

### Custom layout

When nonprofits have a `custom_layout` attribute specified, their campaign pages and peer-to-peer campaign creation page are loaded using that layout. The value of the attribute is a snake-case name of the layout file that exists in `/app/views/nonprofits/custom_campaign_layouts/`. Currently the attribute can only be set via console.

For `safety_around_water` , the custom layout includes:
- a custom banner (1200x120) – currently missing, not yet prepared by YMCA 
- a picture and name, city or state of the campaigner, as well as reason for supporting, if the campaign is peer-to-peer
- slight design changes.

Campaign’s user profile is used to show picture and location. `reason_for_supporting` is an attribute on `Campaign` – perhaps it would be better to extract to a separate object or at least rename.

### Peer-to-peer campaigns & templates

The idea was to allow other people to create campaigns based on a nonprofit campaign, with some limited possibilities to adjust. Currently, campaigners can set their own goal amount, but no other campaign attributes can be modified. The `CampaignTemplate` model holds the information on what the customisable attributes are and values for selected campaign attributes. Templates can be used to create a campaign, sort of like a blueprint. Creating a template uses basically the same form as creating a campaign.

I also introduced a relationship between a parent campaign and a child campaign to keep track of which campaign is peer-to-peer and on whose part it’s supposed to raise money.

## To be done before merge
- [x] make sure the funds from peer-to-peer campaigns make their way to the right nonprofit
- [x] make sure custom campaign page and peer-to-peer campaign creation page look okay on mobile devices
- [x] allow to edit only customisable fields of campaign
- [x] add revised specs (some controller and model specs exist but aren’t very helpful)
- [x] hide banner if banner file missing

## Possible extensions (ideas, not discussed yet)
- ~choose custom layout filename from list~ not, in fact, a good idea
- edit customisable attributes list
- edit templates
- add more campaign attributes to templates for pre-filling
- change banner file for custom layout: DONE – banner can be changed per campaign
- deal with slugs on campaigns created from templates
- create campaigns from other campaigns and inherit their template


ericschultz:
Thanks for the initial work here @idengager! I know you're not done so no worries there.

I'm a little confused about the "custom layout". Are layouts erb files? If so, there's no way that could be uploaded to an SAAS provider. It's not a bad idea for those running their own site though.

ericschultz:
What is the problem that led to YMCA wanting a custom layout?

idengager:
They want to have a campaigner profile included on the peer-to-peer campaign page, some minor CSS changes and a branded banner included on top. The easiest approach was to make a custom ERB layout specifically for them.

I think the feature could be expanded to offer a choice of the campaign layout for each nonprofit (or campaign). The layouts list would be closed, so no uploading needed. For `safety_around_water`, there is nothing really YMCA-specific except the banner image, but with a little work it can become just another campaign attribute. @ericschultz WDYT?

ericschultz:
Thanks for the explanation! I think their requests are super reasonable and in fact I like them! (Although I'd have to consider the CSS changes some). That said, I'm not sure this is the best way to go about doing this.

My biggest worry is that this will lead to a maintenance nightmare over time. If you're a SAAS, you have to make sure that upgrades don't break these ERB files and for dozens or even hundreds of nonprofits it could become a ton of overhead. Additionally, because they're custom to the service, there's no way for upstream to know if we broke something. I can imagine that upgrades would also break nonprofits running their own servers and they won't know why they're broken.

What I'd propose is to handle most of this through settings. I love the idea of having totally flexible layouts  but I think that's something we need to consider in more detail and may require some sort of WYSIWYG editor. For the campaign pages, we could allow people a second banner size, as well as including the campaigner profile on the P2P campaigns. How about you and I set up a time to meet one-on-one to hash through these topics?

Thanks a ton for all the help Kasia. It's really a pleasure working with you.

ericschultz:
As discussed in our call, I added a new copy naming algorithm, called `SlugP2pCampaignNamingAlgorithm` (yes, the second p is lowercase for Rails file loading purposes) and just pushed.

The usecase should work as follows:

```ruby

base_slug = campaign_slug + #slugify the user's name (don't remember method off hand)
algo = SlugP2pCampaignNamingAlgorithm.new(nonprofit_id)
slug = algo.create_copy_name(base_slug)
```

ericschultz:
@idengager: one quick thing I considered. When saving the pics (particularly the user one), you should use the AWS Presigned post feature (https://github.com/houdiniproject/houdini/blob/master/app/controllers/aws_presigned_posts_controller.rb). We need this because the pictures take a really long time to upload and servers often time out before the uploads finish.

This seems to imply we need to handle the carrierwave stuff in the background. Let me think exactly how to do that :)

idengager:
- I removed the ERB file and put the campaigner profile and banner straight into campaign `show` template. Banner is optional, won't show if not added to campaign; the campaigner profile shows up only for peer-to-peer campaigns.
- Campaign `show` container stays at 980px width.
- Campaigns can have `default_reason_for_supporting`, used as a suggestion in peer-to-peer wizard.
- Slugs are nice, but one user (actually, user profile) can host only one child campaign of a given parent.
- Editors of child campaigns can edit only `goal_amount`.

There's definitely room for improvement in terms of code quality, but I'm fairly certain it does what it's supposed to. Two problems remain: async image upload that you mentioned and an issue I'm having with copying over parent campaign images to a new, child campaign. 

I looked at how it's done with [campaign duplication](https://github.com/houdiniproject/houdini/blob/master/lib/insert/insert_duplicate.rb#L32), but it throws S3 `NoSuchKey Error`s for me–perhaps it's an issue with my local setup? I tried assigning `remote_image_url` on the new objects, but it's super slow and timeouts after 1500ms (for good reason). @ericschultz Tips?

app/assets/stylesheets/campaigns/peer_to_peer/page.css.scss
```diff
@@ -11,6 +11,14 @@ body {
 	padding: 0;
 	background: $fog;
 }
+
+body > .campaign-banner {
```
ericschultz: Is the campaign_banner always going to be 980px? Does it (or should it) vary/change size/etc on non-Desktop screens?

app/controllers/campaigns_controller.rb
```diff
@@ -51,7 +54,15 @@ def create
     Time.use_zone(current_nonprofit.timezone || 'UTC') do
       params[:campaign][:end_datetime] = Chronic.parse(params[:campaign][:end_datetime]) if params[:campaign][:end_datetime].present?
     end
-    campaign = current_nonprofit.campaigns.create params[:campaign]
+
+    if !params[:campaign][:parent_campaign_id]
```
ericschultz: Are there checks for situations where the parent_campaign_id is invalid or doesn't belong to the nonprofit the new campaign is being created for?

config/environments/production.rb
```diff
@@ -3,7 +3,7 @@
 	# Settings specified here will take precedence over those in config/application.rb
 
 	# Code is not reloaded between requests
-	config.cache_classes = true
+	config.cache_classes = false
```
ericschultz: We need to cache_classes in production :)

config/environment.rb
```diff
@@ -48,7 +48,7 @@
 
   required(:aws).schema do
     # the region your AWS bucket is in
-    required(:region).filled(:str?)
+    # required(:region).filled(:str?)
```
ericschultz: Why was this done?

client/js/nonprofits/donate/wizard.js
```diff
@@ -191,16 +192,18 @@ const headerDesignation = state => {
...
+  //     steps: [
+  //       {name: I18n.t('nonprofits.donate.amount.label'),   body: amountStep.view(state.amountStep)}
+  //     , {name: I18n.t('nonprofits.donate.info.label'),     body: infoStep.view(state.infoStep)}
+  //     , {name: I18n.t('nonprofits.donate.payment.label'),  body: paymentStep.view(state.paymentStep)}
+  //     ]
+  //   , followup: followupStep.view(state)
+  //   }))
+  // ])
+
+  return h('div.wizard-steps.donation-steps', 'Donation widget should be here.')
```
ericschultz: Why was this done?

app/controllers/campaigns_controller.rb
idengager: The nonprofit_id is deduced from the parent_campaign_id. I added a check for when the parent_campaign_id is invalid.

config/environments/production.rb
idengager: True. 🙂

config/environment.rb
idengager: I reckon region is not strictly needed for S3 (and not for my setup anyway).

client/js/nonprofits/donate/wizard.js
idengager: This was a temporary workaround, I removed it.

app/assets/stylesheets/campaigns/peer_to_peer/page.css.scss
idengager: For desktops and large screens it shouldn't be bigger, but yes: for mobile it should be smaller, accordingly to the whole layout.

idengager:
So, things remaining:

- [ ] check & adjust mobile layout (at least banner size)
- [ ] figure out AWS Presigned post/Carrierwave background upload
- [ ] correctly copying over images from parent to child campaign
- [ ] sync with master.

ericschultz:
Quick question: If a parent changes their images, should it also change the child campaign's images?

My hunch is yes but I don't think that's an immediate need.

ericschultz:
@idengager could you rebase on master? I'm not quite sure what is going on here but I'm having issues getting your version running.

db/structure.sql
```diff
@@ -2,15 +2,14 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 9.6.5
--- Dumped by pg_dump version 9.6.8
+-- Dumped from database version 10.1
```
ericschultz: Please only use database 9.6 for development.

db/structure.sql
```diff
@@ -2,15 +2,14 @@
...
+-- Dumped by pg_dump version 10.1
```
ericschultz: Please only use pg_dump of 9.6. If you don't the dump files don't match and merge chaos ensues.

ericschultz: As an aside, the docker-compose.yml uses the correct versions.

app/controllers/campaigns_controller.rb
```diff
@@ -98,15 +108,21 @@ def name_and_id
 
   def peer_to_peer
     session[:donor_signup_url] = request.env["REQUEST_URI"]
-    @npo = Nonprofit.find_by_id(params[:npo_id])
+    @nonprofit = Nonprofit.find_by_id(params[:npo_id])
+    @parent_campaign = Campaign.find_by_id(params[:campaign_id])
+    @child_campaign = Campaign.where(
+      profile_id: current_user.profile.id,
```
ericschultz: When the user goes here, they'll probably be logged out. If so, the page crashes here as current_user is nil.

ericschultz:
On child campaigns, the user shouldn't be able to create an offsite donation, have a custom email, any gift options, custom receipt, copy campaign. I'm not sure they should even have a dashboard.
child peer-to-peer campaign

ericschultz:
Actually, this brings up an additional question: should child campaigns even have gift options? It's not clear to me that they should.

ericschultz:
Campaign creation image copying works properly now https://github.com/houdiniproject/houdini/pull/48/commits/b890ef7d2f9482cd5185c8e85859bef386f5585c

ericschultz:
There's something wrong with the donation code on the campaign pages. The donation button doesn't even come up. There's an error in the console reading `vnode.data is undefined`.

idengager:
> The donation button doesn't even come up. 

This is an issue I also had with the new campaign wizard 😞It seems like there can't be two instances of the wizard in the views. I dealt with it by introducing a [separate wizard file for creating P2P campaigns](https://github.com/houdiniproject/houdini/pull/48/files#diff-679cf027374828b469a006a8986a76bb), but this is, obviously, more like a workaround that a fix.

My guess is that all the wizard instances save their state globally, so having more than one (even on separate pages) makes a mess. If you have any suggestions on how to fix this for real, please share.

ericschultz:
> Maybe a user that isn't associated with any nonprofit shouldn't even be able to view the campaigns show view? Maybe a simplified index view with a list of campaign link would suffice? If the user clicks on the campaign link, it would render the campaign just as any other user, signed in or not, would see it. @ericschultz Thoughts?

I'd say they should be able to see the Settings since they could want to change the goal amount. Wendy seems to agree with me.

> My guess is that all the wizard instances save their state globally, so having more than one (even on separate pages) makes a mess. If you have any suggestions on how to fix this for real, please share.

Wizard instances don't share their state globally so it couldn't be that. We do need to get the donate button working on the campaign as that's the entire point :)

ericschultz:
>  I dealt with it by introducing a separate wizard file for creating P2P campaigns, but this is, obviously, more like a workaround that a fix.

The fix is moving to React. We have some wizard code already I use for the onboard but it needs to be cleaned up more.

idengager:
> they should be able to see the Settings since they could want to change the goal amount. 

Ah, you are correct of course.

idengager:
Banner uploading is fixed in https://github.com/houdiniproject/houdini/pull/48/commits/2325fe8fc27dae05bce0fcd11094d3f76812b83d and https://github.com/houdiniproject/houdini/pull/48/commits/0854534f9c917a54878f27f0cddbde5a6c77fb23

ericschultz:
@idengager What do you need from me still? What's next steps?

idengager:
@ericschultz I think there's only the presigned post feature for avatar uploads, which I'll look into today.

ericschultz:
One question: How important is it that we get the avatar feature working on this pull request? It's something I would like to consider a bit more and I don't feel I have a clear path forward on finishing it in the next few weeks.

idengager:
@ericschultz The avatar upload feature was expected to increase conversions by adding a personal touch on the peer-to-peer campaigns. Having said that, I guess that time-outs during image upload won't help the conversion rate either.

ericschultz:
Had to do show a demo of this and found a few bugs/polish issues:

* Language on the p2p campaign creation campaign is confusing as it relates to confirmation.
* There should only be the single create campaign button on the campaign pages now. There's one that links to the old method of p2p creation and the new method
* Formatting of the avatar is weird compared to the reason for supporting. It's often way too big and looks funky overall.
* Spacing between avatar, name and city looks off
* City and state should be on same line in the figcaption (below the avatar)

PR #72, opened 2018-06-08
Autocomplete React field for State list

ericschultz:


ericschultz:
We don't use react-autocomplete. This should be changed to react-select instead.

PR #91, opened 2018-08-08
Support for threadsafe mode

ericschultz:
Currently we don't set threadsafe. It's not clear to me that we have any issues because of it now but it prevents us from safely turning on multiple threads in puma which significantly reduces throughput. This pull requests attempts to address this.

I'm a little hesitant to commit this as I'm not totally sure this is safe in all situations yet.


ericschultz:
Notably, this does NOT  turn on multiple threads on puma.

PR #100, opened 2018-10-03
Edit donation via React. Fixes dedication field. Closes #95

### LINKED ISSUE #95: Inconsistent use of dedication field
In the donations database, dedications are handled in two different ways (likely unintentionally)

First if added via the donation wizard, the information is saved into the database as a JSON object. However, when editting the donation from the Edit Donation pane, the dedication is treated as plain text. This causes issues when we assume it is one of the other. In fact, in the edit donation pane, JSON formatted dedications are shown a text representation of the JSON.

In order to fix this inconsistency, I propose the following changes:

* All of the backend-created, plain text dedications should be turned into JSON via a migration
* The edit donation pane should have a section for dedications which allow viewing and editing on the following:
  * The kind of dedication (in honor of, in memory of, not set)
  * The supporter associated with the dedication 
  * The name of the supporter set for the dedication
  * The phone, address and email of the contact for the dedication

While this should be relatively straightforward once we do the React rewrite, we're not there. Therefore, I propose we only allow VIEWING the supporter and name but not editing for the first pass.

ericschultz:
This pull request add the following features: 
* improves the way bootstrap works in Houdini so only a small portion of a page is contained.
* Moves the edit donation form to React.
* Fixes the dedications in the DB so they all are in JSON.


ericschultz:
Fixes needed before merge:
- [ ] remove optional in dedication
- [ ] improve the dedication display on the payments list
- [ ] in the details view of payments, don't link to the dedicatee if they don't have a supporter record.

PR #147, opened 2019-01-30
Remove gmail. Closes #145

### LINKED ISSUE #145: Removal of Gmail access
Recently Google decided to remove the ability of third party applications to access the Gmail API. The Houdini CRM has the ability to manage Gmail threads but it's not a particularly highly used feature. While there are exceptions to the Gmail API restrictions, they're cumbersome. Additionally, the feature could use additional thought. These restrictions go into effect near the end of February.

I propose we remove everything that connects to the Gmail initially. This will keep the Gmail threads in the supporter activity feed but won't allow anyone to interact with them.

In version two, I support moving the entire email supporter feature into a gem so most people don't have to even have it installed. If we decide at some point we want to bring it back, so be it but I don't see a need right now.

ericschultz:


ericschultz:
Waiting to get feedback on linking to a gmail thread before merging.

ericschultz:
I haven't heard any sort of update on this from anyone. The Gmail feature stops working on February 20.

PR #173, opened 2019-03-06
fix docker name so it works consistently

schwindtkevin:


ericschultz:
Seems good to me. I'll commit. Everyone of course has to rebuild now but there are not many of us. :)

PR #323, opened 2020-07-16
Fix the template spacing for component.tsx

wwahammy:


wwahammy:
This appears to be out of date and unneeded.

PR #384, opened 2020-11-12
Various cleanup of statuses

wwahammy:


wwahammy:
Going to merge because this should be passing but isn't.

PR #394, opened 2020-12-14
Add autoupdating of NOTICE-js and NOTICE-ruby

wwahammy:


wwahammy:
This is going to get a bit hacky because we're testing the interaction with dependabot

PR #395, opened 2020-12-15
Add autoupdating of NOTICE-js and NOTICE-ruby

wwahammy:


wwahammy:
Likely to be quite messy because we're figuring things on autoupdating notice files.

PR #449, opened 2021-01-21
Create HoudIDs for stringy ids

wwahammy:
Some of the Houdini event objects will need a stringy id like a UUID. At the same time, it would be nice to have IDs that communicate what type of object it is, like Stripe does. I've create a form of string ID called HoudIDs which consists of a prefix, an underscore and then 22 random alphanumeric characters. (22 is the number you get when converting a UUID to base 62)

In this PR, they are only used for object events and have the format: "objevt_aBc9..."


wwahammy:
If no one comments, I will merge around mid-day January 22, CST

PR #455, opened 2021-02-02
Rename offensive external name for Tag Definitions

wwahammy:
For legacy reasons, we called the definitions of tags at the Nonprofit level a "tag master". This is an offensive term and the Houdini Project apologizes for its use.
This pull request renames "tag master" to tag definition" in all the places related to Object Events. A future PR will need to rename its uses everywhere else.

Once again, we are sincerely sorry for the usage here and will continue to work to make Houdini an inclusive, welcoming FOSS community.


wwahammy:
As this relates to an issue of inclusion and the likely issues caused by the PR would be limited, I will merge as soon as checks pass.

PR #461, opened 2021-02-04
Remove geocoder

wwahammy:
The geocoder is used in a few rare cases but honestly, it's not all that useful and makes everything harder to set up. This PR removes it from the repo and the database.﻿


wwahammy:
Assuming there are no comments, I will merge around mid-day  Feburary 4, CST

PR #464, opened 2021-02-04
Remove unused coupon class

wwahammy:
I've been working in this code for almost four years and never noticed this class before *laughs*

It has no use so we're removing.﻿


wwahammy:
Not going to wait to merge because this is unused code.

PR #465, opened 2021-02-04
Remove unused gmail integration

wwahammy:
Google requires a very expensive ﻿integration fee for integrating with Gmail. This feature has never been very good and just adds a maintenance burden. It was partially removed previously so this finishes the job.


wwahammy:
I intend to merge this immediately since we don't currently use this.

PR #466, opened 2021-02-04
Remove parallel_tests

wwahammy:
Parallel tests were a nice idea but ﻿they don't really work well; the output gets so mucked up that it's basically unusable. There's some work being done at https://github.com/rspec/rspec-rails/issues/2104 and https://github.com/ioquatix/turbo_test but nothing is done yet. We should investigate that in the future but until then, let's remove this code and simplify our lives.


wwahammy:
I would be shocked if this breaks anybody so I'm going to merge immediately.

PR #473, opened 2021-02-08
Moving nonprofit creation to the command line

### LINKED ISSUE #468: [FEATURE] Use a rails command for creating new nonprofits
**Is your feature request related to a problem? Please describe.**
Currently, the registration wizard has a number of limitations. Those include:

* it's written in the legacy Bootstrap and React wizard framework
* it really only makes sense for CommitChange and US nonprofits
* there's no way to restrict nonprofit to admin only.
* It's not easy to get started on Houdini right now.

**Describe the solution you'd like**
While I think long term, it would be good to add a wizard back in, I'm thinking it'd make more sense to move Nonprofit creation to a set of rake tasks. That would give admins the ability to add new nonprofits and users to those nonprofits but not have maintain UIs for now. This would simplify our work immensely. I'll describe what I'm thinking below.

**Describe alternatives you've considered**
We could create a new wizard. But it strikes me as way too much work for minimal gain since we don't totally know the use-cases for it.

NOTE: `add_nonprofit` is a placeholder name. I don't know if this would happen inside a rails generator or a rake task or what exactly.

`add_nonprofit` would be a standard question and answer command line tool and the user would provide the following information:

* nonprofit name
* location (This implicates addressing the routing for state, city and country slugs)
* public email
* public website
* admin user name
* admin user email
* admin user password

We'd also want the user to be able to provide the following info as command line flags:
* whether to automatically confirm this new user
* whether to make them a super-user

Additionally, we should make it possible to provide all of these options as command line arguments so you could automate this (Not sure if it's a necessity though).

wwahammy:
Closes #468﻿

Some of the ideas in https://github.com/rails/rails/pull/39444 are probably useful for this.


wwahammy:
This adds a nonprofit creation command to rails command line. If you run `bin/rails houdini:nonprofit:create`, you will get a set of prompts for creating a new nonprofit. Additionally, if you can pass in arguments to the command for pre-filling the values and some additional options.

As this is a big change, I will leave this PR open until at least mid-day, March 27, CDT for comments before merging.

PR #479, opened 2021-02-09
Move from structure.sql to schema.rb

wwahammy:
One problem we're discovering as we add more people is how difficult it is to keep db/structure.sql accurate. It's so easy to make mistakes with it. The recommendation is usually to use schema.rb unless you simply can't.

After a quick review, it appears that we only had two functions which were preventing us from using schema.rb and they were both unnecessary. This PR removes those functions and switches the project to schema.rb.

I'm going to keep this open to wait for feedback because this is a major change and will likely cause issues with anyone with open PRs.


wwahammy:
I'm going to merge this because it's a massive headache trying to merge structure.sql when you have multiple contributors.

PR #476, opened 2021-02-08
Add campaign_gift.* and campaign_gift_purchase.* events

wwahammy:
Depends on #469

Closes #475
Closes #420

NOTE: While the object event functionality is here, nothing currently fires these events. This requires the rewrite of the transaction creation mechanism.

Currently, setting the campaign gift is a separate API call after the campaign gift purchase which would be known as not great design. Because of that, it's not really possible to separate out campaign gift purchases and donations.

PR #481, opened 2021-02-10
Event publishing for Transaction

### LINKED ISSUE #429: [FEATURE] Transaction events
* transaction.created
* transaction.updated
* transaction.deleted

wwahammy:
This is unlikely to fully complete Transaction work but we'll be firing the events

Depends on #476

Closes #429


spec/lib/insert/insert_donation_spec.rb
```diff
@@ -15,6 +15,7 @@
     describe 'param validation' do
       before(:each) do
         expect(Houdini.event_publisher).to_not receive(:announce).with(:donation_created,any_args)
+        expect(Houdini.event_publisher).to_not receive(:announce).with(:transaction_created,any_args)
```
clarissalimab, line 18: What do you think about moving those expects out from `before` and placing them into `it`s?

wwahammy: I'm fine with that in theory. How would we avoid having to repeat those expects in every it though?

wwahammy:
This is the first pass on Transactions. Assuming there are no comments, we will merge around midday April 13, US Central Time.

PR #491, opened 2021-02-13
Creates ObjectEventListener and WebhookTransmitJob

### LINKED ISSUE #453: [FEATURE] Adding nonprofit specific webhook configurations
# Nonprofit-specific webhooks

As part of the Object Event improvements, it is intended that the nonprofits would be able to set up webhook configurations for given events that happen to that nonprofit. As an example, a nonprofit might want to contact a particular API A when the a supporter is created or when a tag_master is added while they might want to contact a different API B when a supporter is deleted.

## Key terms

A few key terms are useful to understand here.

### Object event

An object event describes when relevant characteristics of a business object have changed. Usually, these apply to when a business object is created, updated or deleted. Currently events are only published after the change has been committed. If you're looking in the `docs/event_definitions` folder, Object Events will be objects of type `HoudiniEvent`. Object events have a type. The type is found by looking at the `object` attribute of a the event as JSON. Its type will follow the format of `object_type.verb`. As an example, an object event of type `supporter.created` would be an object event describing the creation of a supporter.

### Event Publisher

The EventPublisher in Houdini, which is accessed from `Houdini.event_publisher` is a PubSub-like system. Events are published by calling `.announce`. `.announce` takes the following parameters:

* a symbol representing the type of object. This will be the object event type described above but with periods replaced with an underscore. As an example, for an object event type of `supporter.created` you would pass `:supporter_created` here. The listeners on EventPublisher respond when they have a method matching that symbol so it needs to be correct.
* one or more arguments containing the object event. Usually it's just a single argument with the object event as JSON.

Learn more about EventPublisher at `gems/bess/lib/houdini/event_publisher.rb`.

### Listener

A listener is a class that subscribes to events published on `Houdini.event_publisher`. Notably, these subscriptions are based upon
what methods are in the given class. As an example, if the class has a `supporter_created` class method, it will be subscribed
to `supporter.created` events. Make sure to subscribe the class, not an object in `config/initializers/houdini_event_publisher.rb`

## Feature description

This feature will consist of four major parts:

 1. A Webhook Adapter for OpenFn
 2. Model for a ObjectEventHookConfig
 3. Listener to send webhooks to the appropriate service when they occur
 4. Documentation of how to set up an OpenFn job for live testing of the webhook

The parts will be described below.

### A Webhook Adapter for OpenFn

A Webhook Adapter will be a class or set of classes for notifying an external service of an object event as well as any information necessary for calling that service. For now, we only have a Webhook Adapter for OpenFn so let's not get too specific until we have additional experience for using other services.

### Model for an ObjectEventHookConfig

An ObjectEventHookConfig contains the configuration for the following:

* A set of object events types
* The nonprofit that the ObjectEventHookConfig applies to (`belongs_to`/`has_many` relationship)
* The Webhook Adapter describing the service to send the object events to.
* The configuration needed for connecting to the Webhook Adapter. As an example, on OpenFn, each organization on OpenFn has their own
inbox for receiving messages; that would be part of the configuration here. This include the API Key for that inbox (it's sent in the `x-api-key`
header).

### Listener to send webhooks

A listener must be added to `app/listeners` and must inherit from `ApplicationListener`. For each event the listener handles,
the listener should find the appropriate nonprofit, find any applicable `ObjectEventHookConfig`'s for that nonprofit and then use that config
to call the applicable webhook adapters. Make sure that every call to the webhook is in its own ActiveJob call. We want a failure on one specific
webhook call for a single event to not affect the calls to the other webhook calls.

### Documentation of how to set up an OpenFn job for live testing of the webhook

The webhook has to have some live testing. As part of that, there should be some documentation of how the webhook is configured on the OpenFn side
and any code used for that testing. This doesn't need to be a full tutorial but it should be relatively straightforward to follow.

## What isn't in this specific feature?

1. A UI for configuring webhooks for a nonprofit or anyone else
2. Support for other types of webhook filtering (hooks for multiple nonprofits, a subset of events, etc.)

clarissalimab:
Closes #453 

### OpenFn side
1. Navigate to [OpenFn](https://www.openfn.org/signup) and sign up;
2. Set up an authentication method on `Access and Security` option, by clicking on `Add new security protocol`. Choose the API Key auth type;
3. Create a trigger of "Message Trigger" type and define which filter should be in your trigger payload:
image

> In this case, we'll be sending an event of type *supporter_note.created*.

4. Navigate to the project view and copy your inbox link.

#### Creating a job that sends an e-mail using Mailgun

1. Create an account on [Mailgun](https://signup.mailgun.com/new/signup) (for the sake of testing, a free account is enough);
2. For using the sandbox domain Mailgun provides for testing, add your e-mail as an authorized recipient:
image
3. On OpenFn, set up your Mailgun credentials on `Credentials`;
4. Create a new job on `Jobs`, by clicking the `+` icon and connect to the corresponding trigger;
5. Add the script that the job is going to execute on `Execution`. An example for Mailgun could be:

``` javascript
send({
  from: '<your-email>@gmail.com',
  to: '<your-email>@gmail.com',
  subject: 'Supporter note created!',
  text: `A supporter note was created for the supporter ${dataValue('data.object.supporter.name')(state)} of the nonprofit ${dataValue('data.object.nonprofit.name')(state)}.`,
})
```
> To access your payload data from your job, you can use `dataValue('someKey')(state)`

### Application side

``` bash
$ rails c
```

``` ruby
# Set up for the ObjectEventHookConfig
nonprofit = Nonprofit.last
webhook_service = :open_fn
object_event_type = 'supporter_note.created'
configuration = { headers: { 'x-api-key': <your api key> }, webhook_url: <your inbox url> }
attributes = { webhook_service: webhook_service, configuration: configuration, object_event_types: [ object_event_type ] }
object_event_hook_config = nonprofit.object_event_hook_configs.create(attributes)

# Set up for the event
event = {
      'id' => 1,
      'object' => 'object_event',
      'type' => 'supporter_note.created',
      'data' => {
        'object' => {
          'id'=> 1,
          'deleted' => false,
          'content' => 'Some content',
          'nonprofit'=> {
            'id' => nonprofit.id,
            'name' => nonprofit.name,
            'object' => 'nonprofit'
          },
          'object' => 'supporter_note',
          'user' => {
            'id' => 1,
            'object' => 'user'
          },
          'supporter' => {'name' => "Fake Supporter Name"}
        }
      }
    }
Houdini.event_publisher.announce(:supporter_note_created, event)
```

If everything goes well, the OpenFn inbox is going to show a successful run for your job, and on the run details you'll be able to see that the mail sending was enqueued, and in a few instants you'll have received an e-mail.

### DIFF
--- a/app/jobs/webhook_transmit_job.rb
+++ b/app/jobs/webhook_transmit_job.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+class WebhookTransmitJob < ApplicationJob
+  queue_as :default
+
+  def perform(object_event_hook_config, payload)
+    object_event_hook_config.webhook.transmit(payload)
+  end
+end
--- a/app/listeners/object_event_listener.rb
+++ b/app/listeners/object_event_listener.rb
@@ -1,0 +1,158 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+class ObjectEventListener < ApplicationListener
+  def self.campaign_gift_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_option_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_option_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_option_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_purchase_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_purchase_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.campaign_gift_purchase_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.event_discount_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.event_discount_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.event_discount_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_level_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_level_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_level_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_purchase_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_purchase_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.ticket_purchase_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_address_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_address_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_address_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_note_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_note_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.supporter_note_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.donation_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.donation_updated(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.donation_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.custom_field_definition_created(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.custom_field_definition_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  def self.tag_definition_created(event)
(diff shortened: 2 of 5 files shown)


app/jobs/webhook_transmit_job.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+class WebhookTransmitJob < ApplicationJob
+  queue_as :default
+
+  def perform(webhook, payload)
+    webhook.transmit(payload)
+  end
```
clarissalimab, line 10: Not sure if passing the webhook to the job is okay, I can pass the ObjectEventHookConfig id and do a `find_by_id` here

app/listeners/object_event_listener.rb
```diff
@@ -0,0 +1,160 @@
...
+
+  def self.tag_definition_deleted(event)
+    enqueue_transmissions_to_webhooks(event)
+  end
+
+  private
+
+  def self.enqueue_transmissions_to_webhooks(event)
+    object_event_hook_configs(event).each do |config|
+      WebhookTransmitJob.perform_now(config.webhook, event)
```
wwahammy, line 146: Two things: 

* You should enqueue each of these to their own job. Right now, if one of the transmissions fails, then other transmissions fail for that event.
*You should pass the ObjectEventHookConfig to the job itself. ActiveRecord records can be serialized properly for ActiveJobs but regular objects from classes can't.

app/jobs/webhook_transmit_job.rb
wwahammy: See below. The quick answer is 'not really'.

app/listeners/object_event_listener.rb
clarissalimab: What do you mean by their own job? I thought that by making one `perform_now` for each config it would work.

wwahammy: `.perform_now` is the equivalent of saying "run this code right now on the current thread and block until it finishes". `.perform_later` means "queue an ActiveJob onto the proper queue adapter and then don't wait for it to finish". `.perform_now` is synchronous and `perform_later` is async.

clarissalimab: Thanks!

app/listeners/object_event_listener.rb
```diff
@@ -0,0 +1,160 @@
...
+    object_event_hook_configs(event).each do |config|
+      WebhookTransmitJob.perform_later(config, event)
+    end
+  end
+
+  def self.object_event_hook_configs(event)
+    nonprofit(event).object_event_hook_configs.select do |config|
+      config.object_event_types.include? event["type"]
+    end
+  end
```
wwahammy, line 154: It would be good to test this somehow. We want to make sure that there's no chance the wrong webhooks are being called. I'm not quite sure the best way to do that offhand.

wwahammy: Actually, I know how. Move the search for webhook configs into Nonprofit itself. Given `nonprofit` is an instance of `Nonprofit` and `type` is the event type being checked for, which of these APIs feels more natural?

* `nonprofit.find_object_event_hooks_for_type(type)` or,
* `nonprofit.object_event_hooks.for_type(type)`

clarissalimab: I think `nonprofit.object_event_hooks.for_type(type)` sounds better to me!

wwahammy: `nonprofit.object_event_hooks_configs.for_type(type)` it is!

You can add `for_type` as an extension method described at https://guides.rubyonrails.org/association_basics.html#association-extensions

Once you do that, you can unit tests that the method is getting the correct configs

clarissalimab: Actually, I think it would be better if it remained on Nonprofit, so it would return all the corresponding `ObjectEventHooks`. I'd stick with the first option, in that way.

clarissalimab: Ahh, great then!

wwahammy (approved):
I like it! I'll merge around mid-day February 14 CST if no one comments

PR #492, opened 2021-02-15
init_builder can now properly autoexpand enumerable attributes

wwahammy:


wwahammy:
There appears to be a few bugs here so I'll move this to draft

wwahammy:
Made redundant by #565

PR #493, opened 2021-02-16
Fixing Supporter and Supporter Note creation

clarissalimab:
## Supporter creation

The supporter creation controller had a bug where it tried to access a nonexisting value for `nonprofit` on its controller. It caused this problem on the supporter creation modal: 

image

To fix it, I accessed the `current_nonprofit` value instead.

## Supporter note creation

The wrong parameters were being passed to the `InsertSupporterNotes.create` method. It expected a `Supporter`, a `User` and a `content`, from the note to be created. It also had a problem trying to access a symbol that did not exist, as `supporter_params` was a method that returned a string, originally.

### DIFF
--- a/app/controllers/campaigns/supporters_controller.rb
+++ b/app/controllers/campaigns/supporters_controller.rb
@@ -5,7 +5,7 @@
 module Campaigns
   class SupportersController < ApplicationController
     include Controllers::Campaign::Current
-  include Controllers::Campaign::Authorization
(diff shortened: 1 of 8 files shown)


wwahammy:
@clarissalimab thanks for the PR. Please fix the failing specs, we can't merge until they all pass. You get run all of the specs yourself with `bin/rails spec` or specific specs by running `bin/rspec path/to/spec.rb`

PR #495, opened 2021-02-17
Change our ActiveJob backend to good_job

wwahammy:
@clarissalimab and I have been having issues with weird behavior from the EventPublisher. Sometimes events are being handled multiple times. We're not quite sure what's going on so we're going to try using good_job on this PR to see if we can address it. If we can, we'll move fully to good_job.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'sassc', '~> 2.0', '>= 2.0.1'
 gem 'stripe', '~> 1.58' # January 19, 2017 version of the Stripe API https://stripe.com/docs/api
 gem 'webpacker', '~> 5.2.1'
 gem 'react-rails'
+gem 'good_job'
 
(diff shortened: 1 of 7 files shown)


wwahammy:
@clarissalimab next chance you have, try working on this branch and seeing if we can replicate the double firing.

wwahammy:
We should also add the good_job  web monitoring software. To do so, follow the instructions at https://github.com/bensheldon/good_job#dashboard and use the Devise authenticate helper BUT use the `user.super_admin?` method for authentication.

wwahammy:
Since `good_job` is WAY nicer than the built-in ActiveJob code, we'll move to it. I'll merge this around noon  February 24 CST unless someone comments.

wwahammy:
@clarissalimab would you be able to fix that Rspec failure?

wwahammy:
@clarissalimab nevermind actually, it'll be easier for me to do it so we don't have to go through a bunch of merges again

PR #500, opened 2021-02-19
Have Money use BigNumber

wwahammy:
`Money` was previously using number for calculations. This was a problem when you multiply by a fractional amount. This change makes `Money` use `BigNumber` internally so calculations are always accurate.


wwahammy:
This is a necessary change and unlikely to break anything so I will merge immediately.

PR #501, opened 2021-02-21
Add rubocop code analyser

### LINKED ISSUE #488: [FEATURE] Use rubocop!
Currently the project has a less than consistent format for ruby source code. This is not easy to work with. Adding a rubocop format that is part of a build would help. This addition of rubocop must meet the following standards:

* must exclude files which currently are in the codebase. We don't want to do a huge reformat and blow our history to bits. As files have significant rewrites, we can consider adding them to the format requirements.
* must use the tab for tabs. This is an accessibility feature as spaces are less accessible for screenreaders.
* must be included in the build process, i.e. if code doesn't meet the standard, the build should fail.

It also should:

* be somewhat close to one of the formats we use. (and, holy cow, are there a lot of them)
* be automagically sensed by editors like VSCode. If that requires an extension, that extension should be documented if possible.

clarissalimab:
Closes #488.

To see the code analysis locally, run:

```bash
rubocop
```


wwahammy:
The checks seem to be failing for both pronto and the main_build.

.github/workflows/pronto.yml
```diff
@@ -0,0 +1,26 @@
+name: Pronto
+on: ['push', 'pull_request']
+
```
wwahammy, line 3: We should limit this job to only run when ruby related files have changed. Perhaps use the same conditions as main_build uses?

Alternatively, this could be put into the main build.

.rubocop.yml
```diff
@@ -0,0 +1,20 @@
+AllCops:
+  NewCops: enable
+  Exclude:
+    - db/**/*
+    - tmp/**/*
+    - Gemfile
+    - Gemfile.lock
```
wwahammy, line 7: Doesn't this imply that all the old files would be  tested if we ran rubocop?

Gemfile
```diff
@@ -3,18 +3,18 @@
 source 'https://rubygems.org'
 
 ruby '2.6.6'
-gem 'rails', '~> 6.1.3'
-gem 'jbuilder', '~> 2.10'
```
wwahammy, line 7: Make sure not reorder this file, especially things related to rails. Order matters in `Gemfile`s. There's a definitely formatting standards to implement in this file but reordering the imports is not one of them.

wwahammy (changes requested):
Good stuff here but there are some changes you'll need to make.

I'm also not entirely sold on using pronto. Have you used it before?

.rubocop.yml
clarissalimab: Sorry, why? This `Exclude` is ignoring those files.

.github/workflows/pronto.yml
clarissalimab: Great! I'll add to the main_build

clarissalimab:
I have used pronto before. It's not essential but helps with not having to run

`git diff --name-only --no-commit-id $(git merge-base origin/main rubocop) | xargs rubocop --force-exclusion`

because we don't want to run on files that were not changed. And we can use it for reek too if we add it later.

.github/workflows/pronto.yml
clarissalimab: Actually, I think it would be better to run in another job, because we'd probably want to know when both are failing.

wwahammy: In theory, I agree. Unfortunately, it takes a really long time for new jobs to boot as it relates to loading the cache and installing any missing packages. Ideally, everything that can run in parallel should but it's not really practical.

wwahammy:
How much time do we save on each build if we run pronto instead of rubocop?

clarissalimab:
I don't think it saves time, it helps to run the code analysis tools only on the relevant files - which in our case means only running on the modified files instead of in the entire project. Rubocop does not have that configuration, but we could always run only the command I mentioned before to filter changed files between `main` and the current branch.

wwahammy:
I'm still not sure we're on the same page regarding limiting that we run rubocop only on what changed. What's the advantage of that?

wwahammy:
I get it if it saves a bunch of time but if the time difference isn't significant what are we gaining?

clarissalimab:
If we use pronto, we exclude files that currently are in the codebase in the analysis. Rubocop itself does not have a configuration to ignore files that already exist.

If we simply run rubocop, we get:
image

* The first workaround for this would be using pronto:
image
In this case, it checked only the changed lines.

* Or running `git diff --name-only --no-commit-id $(git merge-base origin/main rubocop) | xargs rubocop --force-exclusion`:
image

In this case, it checks the entire file instead of just the lines I've changed, while pronto only checks the lines that I've changed.

That's not our case now, but Pronto would also help if we wanted to add other tools, like `reek` that I mentioned on the issue.

wwahammy:
Couldn't we use exclude in the robocop config to exclude the current files?

wwahammy:
We actually do something similar in ESLint to exclude old files.

wwahammy:
https://www.github.com/houdiniproject/houdini/tree/main/.eslintignore

clarissalimab:
We could, but I thought we'd want to see the warnings if we modified the current files, no? We want rubocop to run only on new files?

wwahammy:
My thinking was we'd only run it on new files and we would slowly add old files over time.

My worries about only running on changes is that it tends to be a pain to set up and I think two people with identical file trees could run it on their systems and their rubocop results could vary. But maybe I'm overthinkinf that last part.

clarissalimab:
No problem, I can remove **pronto** and use only **rubocop** then, but doesn't adding the files slowly take the same effect in the end, except that it's done manually?

wwahammy:
> No problem, I can remove **pronto** and use only **rubocop** then, but doesn't adding the files slowly take the same effect in the end, except that it's done manually?


Pretty much, we get the same result at the end which is getting code formatted to our standard going forward.

clarissalimab:
It's ready for review now :smiley:

.github/workflows/ruby.yml
```diff
@@ -68,4 +68,6 @@ jobs:
           bundle install --jobs 4
       - run: bin/setup ci
       - name: run spec
-        run: bin/rails spec
\ No newline at end of file
+        run: bin/rails spec
+      - name: rubocop
```
wwahammy, line 72: Please reorder these. It's better to have rubocop fail first because it's quicker.

wwahammy (changes requested):
Once minor change and then I think it's ready to go!

wwahammy (approved):
:+1: :fireworks:

PR #509, opened 2021-02-25
Remove simplecov

wwahammy:
We don't actually use simplecov for much so there's not a huge gain to keeping it. We can readd it later if we'd like.


wwahammy:
Haha, I merged another change that does this so we'll close it since it's unnecessary!

Issue #413, opened 2021-01-12
[BUG] `return`, `break` or `throw` in a transaction block is deprecated without replacement


wwahammy:
## Description
When you run `bin/rails spec`, there's a deprecation warning about the use of `return`, `break` or `throw` used in a transaction block. We must remove any such cases as the next version of Rails will always rollback in those cases, even when not preferred.

## Possible Fix
Find every transaction block where we use `return`, `break` or `throw` to leave the block. These blocks will either be associated with a call to `ActiveRecord::Base.transaction` or `Qx.transaction` (which wraps `ActiveRecord::Base.transaction`). I believe we'd want `return`s moved up outside the block to return on the method itself. The `throw`s should be replaced with `raise.

PR #517, opened 2021-03-02
Upgrades ruby to 2.7.2

### LINKED ISSUE #458: [FEATURE] Update to ruby 2.7
Originally, when moving to Rails 6, Houdini didn't use update to ruby 2.7 because there were so many warnings as it relates to the [separation of positional and keyword arguments](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) planned (and released) for Ruby 3. This was in part due to dependencies of Houdini and part our own code. As a number of these issues have been cleaned up in dependencies during the intervening time, it's possible there are fewer dependencies with these warnings.

Might be a good first issue because the contributor probably doesn't need a ton of knowledge about how the project works.

clarissalimab:
Closes #458.


wwahammy:
Did you see any warnings in the log when running specs?

clarissalimab:
Yes, I fixed `DEPRECATED: Use assert_nil if expecting nil. This will fail in Minitest 6`. I've also seen [this warning](https://github.com/houdiniproject/houdini/issues/413) but it's not from the upgrade.

PR #519, opened 2021-03-02
Add spec testing to bess

wwahammy:


wwahammy:
So, the problem I'm having with this PR is I want to run the specs from the `gems/bess` directory. Since it's a separate gem, I want to run tests on it by itself. I've tried a bunch of different way to run it on the build server but all of them seem to have bugs that I don't see when I run locally. It appears this has to do with the way that Bundler works but I'm not versed enough in Bundler to understand what's going on. It may just require some trial an error to get this working.

Another way to handle this might be be to move the Bess gem specs to a separate build action. It's not ideal because you have to waste the time building all the gem dependencies again on that build action but it might work.

clarissalimab:
Why are we ignoring Gemfile.lock from bess?

wwahammy:
My understanding is that gems generally should not have a Gemfile.lock. If
they were strict on versions like that, a user of the gem wouldn't have
flexibility on which versions of dependent libraries the app uses.￼

PR #522, opened 2021-03-06
Removes deprecated 'return's from transaction blocks

### LINKED ISSUE #413: [BUG] `return`, `break` or `throw` in a transaction block is deprecated without replacement

clarissalimab:
Closes #413

### DIFF
--- a/lib/insert/insert_duplicate.rb
+++ b/lib/insert/insert_duplicate.rb
@@ -17,9 +17,8 @@ module InsertDuplicate
       raise ParamValidation::ValidationError.new("#{profile_id} is not a valid profile", key: :profile_id)
     end
 
+    dupe = campaign.dup
     Qx.transaction do
-      dupe = campaign.dup
-
       dupe.slug = SlugCopyNamingAlgorithm.new(Campaign, dupe.nonprofit.id).create_copy_name(dupe.slug)
       dupe.name = NameCopyNamingAlgorithm.new(Campaign, dupe.nonprofit.id).create_copy_name(dupe.name)
       if dupe.end_datetime && dupe.end_datetime.ago(7.days) < DateTime.now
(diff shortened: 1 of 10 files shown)


wwahammy:
Looks like you fixed it!

lib/insert/insert_duplicate.rb
```diff
@@ -17,9 +17,8 @@ def self.campaign(campaign_id, profile_id)
       raise ParamValidation::ValidationError.new("#{profile_id} is not a valid profile", key: :profile_id)
     end
 
+    dupe = campaign.dup
     Qx.transaction do
-      dupe = campaign.dup
-
```
wwahammy, line 22: I just realized, I think campaign.dup should happen inside the transaction block. I think if you change line 20 to `dupe=nil`, this should work just fine.

lib/insert/insert_duplicate.rb
```diff
@@ -55,9 +53,8 @@ def self.event(event_id, profile_id)
       raise ParamValidation::ValidationError.new("#{profile_id} is not a valid profile", key: :profile_id)
     end
 
+    dupe = event.dup
```
wwahammy, line 56: Same comment as line 20.

lib/update/update_donation.rb
```diff
@@ -70,9 +70,9 @@ def self.update_payment(donation_id, data)
       end
     end
 
+    donation = existing_payment.donation
+    ret = donation.attributes
```
wwahammy, line 74: Maybe move this one inside the transaction block too. I'm not positive it really matters though.

wwahammy:
@clarissalimab I had three comments here. If those fixes make sense, please submit a new pull request implementing them.

PR #553, opened 2021-03-27
Updates Nonprofit's website validation

### LINKED ISSUE #457: [BUG] Nonprofit URL validation
## Description
On the nonprofit register screen, starting from the onboard, the user is being required to type the "https://www." prefix on the URL field.

## Expected Behavior
It should not be mandatory to write the "https://www." prefix on the URL field for the nonprofit registering, or, at least, the user should receive meaningful feedback that they must fill the field with the prefix.

## Actual Behavior
The user can't register the nonprofit if they fill the URL field only with the "example.com" data, but the user account gets registered. The user also does not get any feedback about what happened.

## Possible Fix
Update the validation to require the "https://www." URL prefix or remove that requirement from the backend.

## Steps to Reproduce

1. Set up the environment for the first time;
2. Run the server and navigate to `localhost:5000/onboard`;
3. Fill the registration form. On the URL field for the nonprofit, fill in with a website that does not contain the "https://www." prefix.

## Context
I was registering for the first time in my development environment.

## Your Environment

* Ruby version: ruby 2.6.6
* Operating System and version: Arch (kernel 5.9.14-arch1-1)

clarissalimab:
Resolves #457

### DIFF
--- a/app/models/nonprofit.rb
+++ b/app/models/nonprofit.rb
@@ -101,7 +101,7 @@ class Nonprofit < ApplicationRecord
   validates_format_of :email,  with: Email::Regex, allow_nil: true
   validates_presence_of :slug
   validates_uniqueness_of :slug, scope: %i[city_slug state_code_slug]
-  validates :website, url: { schemes: ['http', 'https'], public_suffix: true }, allow_nil: true
+  validates :website, url: { schemes: ['http', 'https'], public_suffix: true, no_local: true }, allow_nil: true
 
   validates_presence_of :user_id, on: :create, unless: -> {register_np_only}
   validate :user_is_valid, on: :create, unless: -> {register_np_only}
@@ -144,6 +144,7 @@ class Nonprofit < ApplicationRecord
     set_slugs
     set_user
     add_billing_subscription
+    add_scheme_to_website
     self
   end
 
@@ -274,6 +275,22 @@ private
   def add_billing_subscription
     billing_plan = BillingPlan.find(Houdini.default_bp)
     b_sub = build_billing_subscription(billing_plan: billing_plan, status: 'active')
+  end
+
+  def add_scheme_to_website
+    website = self.website
+    begin
+      uri = URI.parse(website)
+      host = uri && uri.host
+      scheme = uri && uri.scheme
+
+      valid_scheme = host && scheme
+      valid_website = website && website.include?('.')
+
+      self.website = 'http://' + website if scheme.blank? && valid_website
+    rescue URI::InvalidURIError => e
+      e
+    end
   end
 
   def user_registerable_as_admin
--- a/app/models/nonprofit.rb
+++ b/app/models/nonprofit.rb
@@ -99,9 +99,9 @@ class Nonprofit < ApplicationRecord
   validates :city, presence: true
   validates :state_code, presence: true
   validates_format_of :email,  with: Email::Regex, allow_nil: true
-  validates_format_of :website, with: URI.regexp(['https', 'http']), allow_nil: true
   validates_presence_of :slug
   validates_uniqueness_of :slug, scope: %i[city_slug state_code_slug]
+  validates :website, url: { schemes: ['http', 'https'], public_suffix: true }, allow_nil: true
 
   validates_presence_of :user_id, on: :create, unless: -> {register_np_only}
   validate :user_is_valid, on: :create, unless: -> {register_np_only}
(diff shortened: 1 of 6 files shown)


wwahammy:
Oops, I should have clarified. It should still validate but if the url doesn't have a scheme, we should add http. The key is you should be able to not provide a scheme and still have it work.

wwahammy:
Now that I look at this, maybe it'd make sense to use something like this: https://github.com/perfectline/validates_url? It would make sure we have the correct. I'm going to test it quick.

clarissalimab:
@wwahammy I've updated it, it is ready for review now!

app/models/nonprofit.rb
```diff
@@ -276,6 +277,13 @@ def add_billing_subscription
     b_sub = build_billing_subscription(billing_plan: billing_plan, status: 'active')
   end
 
+  def add_scheme_to_website
+    website = self.website
+    if website.present? && (website =~ URI.regexp(['https', 'http'])).blank?
```
wwahammy, line 282: What happens if they have an non `https` or `http` scheme? I think this should only modify it if the scheme is missing.

spec/models/nonprofit_spec.rb
```diff
@@ -143,11 +143,22 @@
         expect(nonprofit_with_bad_email_and_website.errors['email'].first).to match /.*invalid.*/ 
       end
 
-      it 'marks website as having errors if they do' do
-        expect(nonprofit_with_bad_email_and_website.errors['website'].first)
-          .to match('is not a valid URL')
+      context 'when the website has errors' do
+        it 'marks as having weeoes if they don not have a public suffix' do
```
wwahammy, line 147: What's the title of this spec? :smile:

wwahammy (changes requested):
Only a few changes needed here and we'll be good to go!

spec/models/nonprofit_spec.rb
clarissalimab: hahaha it should be 'errors' instead of weeoes, I don't know what happened :laughing:

app/models/nonprofit.rb
clarissalimab: Fixed!

app/models/nonprofit.rb
```diff
@@ -276,6 +277,13 @@ def add_billing_subscription
     b_sub = build_billing_subscription(billing_plan: billing_plan, status: 'active')
   end
 
+  def add_scheme_to_website
+    website = self.website
+    if website.present? && URI.parse(website).scheme.blank?
```
wwahammy, line 282: If you pass in a totally invalid URI, a URI::InvalidURIError will be raised. Like what happens if they pass an array? or an invalid string? Or a hash? 

The validation should be catching bad URL's. You may want to refer to or reuse the code in `validate_url`'s source to reduce the chance of forgetting something. (Although totally possible they haven't thought of everything as well.)

wwahammy (changes requested):
I think there's just the one change.

Also, as you've changed the set of gems, you should run `bin/rails notice:ruby:update`. That updates our NOTICE-ruby file.

app/models/nonprofit.rb
clarissalimab: I used just some parts of it regarding the scheme to do nothing if it is valid, the rest would be duplicated with the validation itself. What do you think?

spec/models/nonprofit_spec.rb
```diff
@@ -143,8 +143,27 @@
...
+            .to eq('http://a_website.com')
+        end
+
+        it 'marks as having errors and does not replace the scheme if it is different than http or https' do
+          nonprofit_with_bad_email_and_website.update(website: 'ftp://invalid.com')
+            expect(nonprofit_with_bad_email_and_website.errors['website'].first)
+              .to match('is not a valid URL')
+            expect(nonprofit_with_bad_email_and_website.website)
+              .to eq('ftp://invalid.com')
+        end
```
wwahammy, line 166: Make sure to check the weird situations we need to change the `add_scheme_to_website` for. they include website being:

* an array
* a string with a space
* a number
* a hash

Maybe make a describe section just for that testing that?

wwahammy: Or combine the sections here, I don't think it matters a ton.

wwahammy (changes requested):
Let's unit test `add_scheme_to_website` to make sure it doesn't blow up.

wwahammy:
As long as this passes, it looks good!

wwahammy (approved):
Looks good!

PR #566, opened 2021-04-12
Add offline_transaction_refund to object events

wwahammy:
Depends on #481
Adding a refund mechanism to offline transactions


wwahammy:
Assuming there are no comments, this feature will be merged around mid-day June 18, US Central time.

PR #567, opened 2021-04-12
Event publisher support for offline_transaction_dispute.* events

wwahammy:
It's possible for offline transactions to have disputes (although unlikely).

This is the starting point for actually adding offline_transaction_dispute.* events.


wwahammy:
Assuming there are no comments, I plan to merge this around mid-day June 22, US Central Time

PR #569, opened 2021-04-13
Fix #1

### LINKED ISSUE #1: Two rspec tests fail when run in suite but not individually
There are two rspec tests which fail when run as part of the spec suite `rake spec`

Add more detail :)

nalight:
Fixes #1


wwahammy:
This is clearly spam. You will be blocked from using this repo and reported.

PR #582, opened 2021-04-22
Move info about the hoster into its own module

wwahammy:
This makes it possible to get information about the hoster of an Houdini instance. This is accessed via Houdini.hoster.


wwahammy:
If there are no comments, I will merge around midday April 23, US Central Time.

PR #587, opened 2021-05-04
Replaces 'commitchange' prefix with 'houdini' from databases

### LINKED ISSUE #124: Default database names shouldn't be prefixed with `commitchange`
For historical reasons, the default database names are prefixed with `commitchange_`. Fixing this requires not just changing the database configuration in rails but also making sure no code, such as scripts, refer to the original database names.

clarissalimab:
Resolves #124.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -16,7 +16,7 @@
 development:
   adapter: postgresql
   encoding: unicode
-  database: commitchange_development
+  database: houdini_development
   pool: 20
   username: houdini_user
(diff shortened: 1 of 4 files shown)


clarissalimab:
Well remembered! I think one week is long enough, @wwahammy

wwahammy:
As announced on Zulip, assuming there's no feedback, we'll be merging this PR near mid-day May 11 US Central time.

PR #610, opened 2021-05-18
Replace Roadie with Premailer

wwahammy:
Roadie has been the email style inliner for Houdini. Unfortunately, Roadie is no longer actively supported and it has a bug preventing using a CDN for email related assets. Premailer is well supported and has no such limitations.

CommitChange has moved to using Premailer and it's very functional.


wwahammy:
I don't think there's much controversy here so, unless someone comments by around mid-day May 19, US Central Time, I'll merge.

PR #612, opened 2021-05-20
Update to webpacker 5.4.0

wwahammy:


wwahammy:
Can't update the NOTICE files because ClearlyDefined is down. I will when it's back up.

PR #613, opened 2021-05-21
Validates nonprofit's timezone

clarissalimab:
Validates that the time zones used by the nonprofits are the ones that come from `ActiveSupport::TimeZone`, which are also compatible with Postgresql.


wwahammy:
Thanks @clarissalimab. I don't see any reason to not merge this so we will!

PR #616, opened 2021-05-25
Consider timezone on supporters and payments queries

clarissalimab:


wwahammy (approved):
This makes sense to me. I'll merge

PR #623, opened 2021-07-05
Includes documentation for FriendlyURLs

### LINKED ISSUE #595: [BUG] Lack of documentation for Friendly URLs
Houdini has a concept which I'm going to call Friendly URLs for nonprofits, campaigns and event pages. To go to one of these pages, you can use a standard Rails path scheme like `/nonprofits/1` or you can use a URL like `/wi/appleton/fox-valley-poverty`. This need to be documented because they're causing issues as described in #586 and people don't understand why.

@wwahammy  needs to document this.

clarissalimab:
Closes #595.

### DIFF
--- a/docs/friendly_urls.md
+++ b/docs/friendly_urls.md
@@ -1,0 +1,31 @@
+# Friendly URLs
+
+Houdini has a concept called "Friendly URLs", that is an alternative for the
+default Rails URLs to present the same navigation in a more user-friendly way.
+
+The base for the URLs start with the nonprofit's state code, city and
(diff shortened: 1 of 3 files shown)


docs/friendly_urls.md
```diff
@@ -0,0 +1,22 @@
+# Friendly URLs
```
clarissalimab, line 1: Is there any other detail that I should add?

wwahammy: No, I like this.

wwahammy (approved):
I think this is good!

PR #638, opened 2021-08-04
Add api/users/current

wwahammy:
Creates a REST api for ﻿getting the current logged-in user. Useful for checking whether the current user is logged in.

### DIFF
--- a/app/views/api/users/current.json.jbuilder
+++ b/app/views/api/users/current.json.jbuilder
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+json.partial! '/api/users/user', user: @user
--- a/app/controllers/api/users_controller.rb
+++ b/app/controllers/api/users_controller.rb
@@ -7,7 +7,9 @@ class Api::UsersController < Api::ApiController
 
 	before_action :authenticate_user_with_json!
 
+	# Returns the current user as JSON
+	# If not logged in, causes a 401 error
 	def current
-		render locals: { user: current_user }
+		@user = current_user
 	end
 end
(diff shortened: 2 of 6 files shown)


app/views/api/users/current.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+json.partial! '/api/users/user', user: @user
```
clarissalimab, line 5: Why did you choose this design? I'm not sure if I understand it, could you tell me more about it?

wwahammy: I'm trying to follow best practices for Rails and Jbuilder which seems to be as follows:

* use as simple of a jbuilder template as possible for the action template,
* accept the relevant records as a class variable
* pass the class variable to a jbuilder partial.


`json.partial! '/api/users/user', user: @user` renders the partial at `app/views/api/users/_user.json.jbuilder` and passes in a local `user` variable set to `@user`.

One advantage of this design is you can use view specs. For example: `spec/views/api/users/current.json.jbuilder_spec.rb` allows us to assign the `@user` class variable at https://github.com/houdiniproject/houdini/pull/638/files#diff-48e80d7ab45ecabafe0b860e0dc3f1fca65877b9df14cdadd8c0a54a32fd9f6dR9 and then verify that the rendered JSON is correct. By doing that, we don't have to worry about logging in properly or making the correct `get` calls to the route that links to the controller.

We can then simply test the controller using `spec/requests/api/users_controller_spec.rb` for everything related to accepting data and handling responses but don't have to test all of the details of the output (although we could!)

clarissalimab: That's very cool! Thanks :)

clarissalimab (approved):
Awesome work! :rocket:

wwahammy:
I don't see anyone complaining about this so I'm going to merge!

PR #639, opened 2021-08-06
Change Unauthorized response format based on the respond_to format

wwahammy:
Currently, we have to manually set whether we want to provide a JSON or HTML response on a controller action when a user is authorized. This change makes it so the response is based on whether the client has asked for JSON or not. Less configuration!

Depends on #638


clarissalimab (approved):
Nice work! :+1:

By the way, I think it's a good idea for us to start removing the ignored files from rubocop as we work on them, so in that way we won't be "hurting" the history that much.

wwahammy:
> Nice work! :+1:
> 
> By the way, I think it's a good idea for us to start removing the ignored files from rubocop as we work on them, so in that way we won't be "hurting" the history that much.

Thank you!

I totally agree on rubocop! Are any of the files I worked on in the ignore? I'd check but I'm on my phone 🙂

clarissalimab:
Yes, those are being ignored:
- 'app/controllers/api/api_controller.rb'
- 'app/controllers/direct_uploads_controller.rb'
- 'spec/controllers/api/nonprofits_controller_spec.rb'
- 'spec/factories/users.rb'

PR #640, opened 2021-08-06
Add api/nonprofits/:nonprofit_id/supporters and api/nonprofits/:nonprofit_id/supporters/:id

wwahammy:
Depends on #639

### DIFF
--- a/spec/requests/api/supporters_request_spec.rb
+++ b/spec/requests/api/supporters_request_spec.rb
@@ -1,0 +1,168 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+require 'rails_helper'
+
+RSpec.describe 'Api::SupportersController', type: :request do
+	let(:supporter) { create(:supporter_with_fv_poverty) }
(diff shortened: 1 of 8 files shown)


spec/requests/api/supporters_request_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+			sign_in user
+			get "/api/nonprofits/#{nonprofit.id}/supporters/#{supporter.id}"
+			expect(response).to have_http_status(:success)
+		end
+
+		it 'returns http success when not logged in' do
+			get "/api/nonprofits/#{nonprofit.id}/supporters/#{supporter.id}"
+			expect(response).to have_http_status(:unauthorized)
+		end
+	end
```
clarissalimab, line 43: Wouldn't it be a good idea to test how the return looks like? I think it also helps as a documentation for the API.

clarissalimab (commented):
Just a :memo:

spec/requests/api/supporters_request_spec.rb
wwahammy: Probably.

There's multiple schools of thought in the Rails community on that but it wouldn't hurt to do that. :+1:

wwahammy: I've added this to the specs!

PR #649, opened 2021-08-12
Add /api/nonprofit/:nonprofit_id/custom_field_definitions and /:id

### LINKED ISSUE #645: [FEATURE] Add api/nonprofits/:nonprofit_id/custom_field_definitions/ and ../:id

wwahammy:
Closes #645﻿

### DIFF
--- a/app/views/api/custom_field_definitions/index.json.jbuilder
+++ b/app/views/api/custom_field_definitions/index.json.jbuilder
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+json.array!(
+	@custom_field_definitions,
+	partial: '/api/custom_field_definitions/custom_field_definition',
+	as: 'custom_field_definition'
+)
--- a/spec/factories/custom_field_masters.rb
+++ b/spec/factories/custom_field_masters.rb
@@ -3,8 +3,12 @@
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
 FactoryBot.define do
-  factory :custom_field_master do
-    nonprofit { '' }
-    name { 'MyString' }
-  end
+	factory :custom_field_master do
+		nonprofit { '' }
+		name { 'Def Name' }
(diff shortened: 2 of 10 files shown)


spec/factories/custom_field_masters.rb
```diff
@@ -5,6 +5,12 @@
...
     nonprofit { '' }
-    name { 'MyString' }
+    name { 'Def Name' }
+    factory :custom_field_definition_with_nonprofit do 
+      name { 'Def Name' }
+      nonprofit { create(:fv_poverty)}
+    end
   end
+
+
```
clarissalimab, line 15: I think this file could be removed from rubocop ignored files!

app/views/api/custom_field_definitions/index.json.jbuilder
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
+json.array! @custom_field_definitions,
+												partial: '/api/custom_field_definitions/custom_field_definition',
+												as: 'custom_field_definition'
```
clarissalimab, line 7: I find it interesting that rubocop thinks this format looks better.

clarissalimab (approved):
:rocket:! But, a reminder and a comment:

app/views/api/custom_field_definitions/index.json.jbuilder
wwahammy: Ya, I hate it. I think the problem is rubocop is really optimized for spaces instead of tabs. Part of me wonders if we should just move to spaces to be like the rest of the Ruby community.

clarissalimab: I think it would work to manually do:
```suggestion
json.array!(
	@custom_field_definitions,
	partial: '/api/custom_field_definitions/custom_field_definition',
	as: 'custom_field_definition'
)
```
instead of accepting its automatic auto-correct. One thing I noticed about using tabs is that it looks somehow bad when reviewing on Github, but not terrible.

spec/factories/custom_field_masters.rb
wwahammy: Done!

app/views/api/custom_field_definitions/index.json.jbuilder
wwahammy: Done in latest push!

PR #658, opened 2021-08-15
Block the creation of a nonprofit when using a non-US state

### LINKED ISSUE #586: [BUG] The nonprofit creation allows you to create it with a state that is not in the US
## Description

When using the nonprofit creation command (`bin/rails houdini:nonprofit:create`), I was allowed to create it with a state that is not from the US, but when I navigated to its dashboard, I had an error:

image

## Expected Behavior

A validation error should come up if the user tries to create a nonprofit that is not from the US.

## Steps to Reproduce

1. Start the project from scratch;
2. Create the first nonprofit by running `bin/rails houdini:nonprofit:create` with the following options:
```sh
What is the nonprofit's name? Dognity
What is the nonprofit's state? DF
What is the nonprofit's city? Brasília
[OPTIONAL] What is the nonprofit's public website? 
[OPTIONAL] What is the nonprofit's public e-mail? 
[OPTIONAL] What is your nonprofit's public phone number? 
What is your nonprofit's admin's name? Clarissa
What is your nonprofit's admin's email address? (It'll be used for logging in) clarissa@commitchange.com
What is the nonprofit's admin's password?
```
3. Start the server, navigate to `localhost:5000` and log in;
4. You should see the error instead of the dashboard.

marcelog5:
Closes #586

# Nonprofit creation have a validation to permitt only US state.

## Description
When try to create a nonprofit that is not in US a error come up.

[[BUG] The nonprofit creation allows you to create it with a state that is not in the US](https://github.com/houdiniproject/houdini/issues/586)

## Acceptance criteria

1. [X] Nonprofit are only created when use states in US?
2. [X] Nonprofit creation are blocked when use a state that is not in US?
3. [X] Tests are working?


app/models/nonprofit.rb
```diff
@@ -107,6 +107,7 @@ def build_admin(**kwargs)
   validate :user_is_valid, on: :create, unless: -> {register_np_only}
   validate :user_registerable_as_admin, on: :create, unless: -> {register_np_only}
   validate :timezone_is_valid
+  validate :state_in_us?, on: %i(create update)
```
wwahammy, line 110: There's no real gain setting the "on" argument to `create` and `update` so it should be removed. The default is to validate before every save and the two "types" of saves are `create` and `update` so mentioning them both is redundant.

config/state_code_constraint.rb
```diff
@@ -3,12 +3,7 @@
...
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/master/LICENSE
 class StateCodeConstraint
-  US_STATES = %w(AL AK AZ AR CA CO CT DE DC FL GA HI ID IL IN 
-      IA KS KY LA ME MD MA MI MN MS MO MT NE NV NH NJ 
-      NM NY NC ND OH OK OR PA PR RI SC SD TN TX UT VT VA WA WV WI WY)
-  US_STATES_REGEX = Regexp.union(*US_STATES.map{|i| /^#{i.downcase}$/})
-
   def matches?(request)
-    US_STATES_REGEX =~ request.params[:state_code]
+    ISO3166::Country[:US].subdivisions[request.params[:state_code].upcase] != nil
```
wwahammy, line 7: I like this idea of using a gem here!

Assuming that subdivision is a `Hash` or a subclass of `Hash` (which I think it is), there's a better way to check if a key exists. `Hash` has a method called `has_key?` which accepts the key you want to search for as the first parameter.

That would change this to: `!ISO3166::Country[:US].subdivisions.has_key? request.params[:state_code].upcase`

Does this work?

app/models/nonprofit.rb
```diff
@@ -180,6 +181,10 @@ def as_json(options = {})
     h
   end
 
+  def state_in_us?
```
wwahammy, line 184: Two things on this line:

* Methods shouldn't end with a question mark in Ruby unless they return a boolean. Validation methods for models don't return a boolean, therefore, you should remove the question mark.

* I'm not sure this is the best name for this method. It's not wrong exactly but if we generalize the set of state codes allowed, we'll have to rename it. Is there a name that makes sense more generally?

app/models/nonprofit.rb
```diff
@@ -180,6 +181,10 @@ def as_json(options = {})
...
+    errors.add(:state_not_in_us, 'Only US states are allowed') unless ISO3166::Country[:US].subdivisions[state_code]
```
wwahammy, line 185: When you add the errors on validation, you're often validating the attributes which have been passed in. In this case, you're validating `state_code`. The way to indicate that in ActiveRecord errors is to pass the attribute as a symbol to the first parameter, i.e. pass `:state_code`.

Additionally, as you noticed in your test, Rails displays errors as a phrase. Therefore, the second parameter needs to be formatted as the rest of the phrase.

A good phrase for this might be "State code must be a US two-letter state code".


One last thing: I think you can use `has_key?` to check the `state_code` just like I mention in the comment down in `config/state_code_constraint.rb`. That would be a better choice.

spec/controllers/api/nonprofits_controller_spec.rb
```diff
@@ -27,7 +27,7 @@
 			input = {}
 			post '/api/nonprofits', params: input, xhr: true
 			expect(response).to have_http_status :unprocessable_entity
-			expect(response.parsed_body['errors'].keys).to match_array %w[name city state_code slug user_id]
+			expect(response.parsed_body['errors'].keys).to match_array %w[name city state_code slug user_id state_not_in_us]
```
wwahammy, line 30: Once changing the errors in `app/models/nonprofit.rb` this change won't be needed.

wwahammy (changes requested):
Thank you for your first submission to Houdini Project! This is some really good work and I think you're not far from completion. I've requested a few tweaks. If you have any questions about them, please reply to the comments.

Also, when you've pushed the version you'd like re-reviewed, please request re-review in the PR UI so I know it's ready for me to check out.

Great work so far, I'm really liking what I'm seeing!

app/models/nonprofit.rb
thiagolopess: About the "?",  we just mistyped it. Thanks for the correction!

Regarding the name of the method, what about replacing it to "is_state_valid" or "state_is_valid" so it doesn't look like a question (which would infer a boolean return type)? This way we think it would be more general.

app/models/nonprofit.rb
thiagolopess: Okay, we'll fix it. Thanks for the explanation!

app/models/nonprofit.rb
wwahammy: I think `state_is_valid` would be good. Let's go with that!

config/state_code_constraint.rb
thiagolopess: We swapped for it and it works! Thanks!

app/models/nonprofit.rb
thiagolopess: We agree on it, much appreciated!

spec/controllers/api/nonprofits_controller_spec.rb
thiagolopess: Indeed.

app/models/nonprofit.rb
```diff
@@ -180,6 +181,10 @@ def as_json(options = {})
     h
   end
 
+  def state_is_valid
+    errors.add(:state_code, 'must be a US two-letter state code') unless ISO3166::Country[:US].subdivisions.has_key? state_code
```
wwahammy, line 185: You should probably upcase the state_code like you do in `config/state_code_constraints.rb`

spec/models/nonprofit_spec.rb
```diff
@@ -121,7 +124,11 @@
         expect(nonprofit.errors['state_code'].first).to match /.*blank.*/
       end
 
-      it 'rejects an invalid user' do 
+      it 'has an error for not in the US state' do
+        expect(nonprofit.errors['state_code']).to include 'must be a US two-letter state code'
+      end
```
wwahammy, line 129: Make sure that no error is added for a state_code if state_code is set to a non-capitalized version of the state_code. As an example, if you set state_code to "Wi" then the validation should pass.

wwahammy (changes requested):
Looking very good! A few minor tweaks and it'll be ready to go!

app/models/nonprofit.rb
igor-paiva: You're right, the upcase method should be used. But in order to use it, it's necessary to check if the `state_code` is not nil.

Something like this:

```rb
  def state_is_valid
    if state_code && !ISO3166::Country[:US].subdivisions.has_key?(state_code.upcase)
      errors.add(:state_code, 'must be a US two-letter state code')
    end
  end
```

spec/models/nonprofit_spec.rb
igor-paiva: Okay, it's a good test case, we'll add it!

While investigating it I've found another problem. We created a `nonprofit_with_not_US_state` but we didn't use it in any test case, a new test case should also be added to cover this case.

Changing the validation method to validate the state code only when it's not nil (as mentioned in another comment), it could look like this:

```rb
  let(:nonprofit_with_not_US_state) { Nonprofit.new(user_id: user.id, state_code: 'KK')}

  it 'has an error for not in the US state' do
     expect(nonprofit_with_not_US_state.errors['state_code']).to match_array ['must be a US two-letter state code']
  end
```

Because it should only have a single error.

app/models/nonprofit.rb
wwahammy: You're on the right track! Ruby actually has a nil-safe navigation operator `&.` which, if you the object on the left is nil, then the whole expression becomes nil. This allows us to keep the search this is a one-liner as follows:

```ruby
errors.add(:state_code, 'must be a US two-letter state code') unless ISO3166::Country[:US].subdivisions.has_key? state_code&.upcase
```

app/models/nonprofit.rb
```diff
@@ -82,7 +82,7 @@ def build_admin(**kwargs)
   end
   has_many :tag_masters, dependent: :destroy
   has_many :custom_field_masters, dependent: :destroy
-
```
wwahammy, line 85: Why all these formatting changes?

wwahammy (changes requested):
I think this is looking good. I only had one more concern. Could you create a [routing spec](https://relishapp.com/rspec/rspec-rails/docs/routing-specs) to verify that the state code actually works as part of routing?

The idea is that we want to know that, if a Nonprofit has a state_code_slug of "wi", a city_slug  of  "appleton" and a name_slug of "fox-valley-fighting-poverty", that a get call to "/wi/appleton/fox-valley-fighting-poverty" routes to the show action on the NonprofitController and `/WI/appleton/fox-valley-fighting-poverty` does as well but `/FR/appleton/fox-valley-fighting-poverty` doesn't becuase FR is not a valid US state code.

I hope that request makes sense. If not,let me know!

app/models/nonprofit.rb
igor-paiva: I'm sorry about that, I didn't mean to change it. I think somehow I've activated an auto formatter. I'll fix this.

thiagolopess: About the test, I think it's a great idea! We'll be adding it.

thiagolopess: Sorry for the delay, we were kinda busy these days.

wwahammy: No need to apologize! Your contributions are super valuable and we're thankful you're willing to volunteer and help out!

wwahammy (approved):
Looks good to me!

wwahammy:
I'm going to rebase and squash all of the commits into one commit and then merge. Thanks so much for this @marcelog5 @igor-paiva and @thiagolopess!

PR #662, opened 2021-08-17
Add /api/nonprofits/:nonprofit_id/campaigns/:campaign_id/campaign_gift_options and ./:id

### LINKED ISSUE #659: [FEATURE] Add api/nonprofits/:nonprofit_id/campaigns/:campaign_id/campaign_gift_options and ./:id

wwahammy:
Closes #659
﻿

### DIFF
--- a/app/models/campaign_gift_option.rb
+++ b/app/models/campaign_gift_option.rb
@@ -31,7 +31,27 @@ class CampaignGiftOption < ApplicationRecord
   after_destroy_commit :publish_deleted
 
   has_one :nonprofit, through: :campaign
+  delegate :currency, to: :nonprofit
 
+  def gift_option_amounts
+    output = []
+    if amount_one_time
+      output.push(GiftOptionAmount.new(
+        Amount.new(amount_one_time, currency)
+      ))
+    end
+    if amount_recurring
+      output.push(GiftOptionAmount.new(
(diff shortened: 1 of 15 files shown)


app/models/campaign_gift_option.rb
```diff
@@ -78,15 +75,19 @@ def to_builder(*expand)
...
 
-      json.gift_option_amount gift_option_amount do |desc|
+      json.gift_option_amount gift_option_amounts do |desc|
         json.amount do 
-          json.currency desc[:amount][:currency]
-          json.cents desc[:amount][:cents]
+          json.currency desc.amount.currency
+          json.cents desc.amount.cents
+        end
+        if desc.recurrence
```
clarissalimab, line 83: Is there any chance that this comes as an empty string instead of `nil`?

wwahammy: Hmmm, I'm not sure I understand what you're asking.

clarissalimab: If `desc.recurrence` comes as an empty string in that verification, that condition is going to be `true` and something will break in the next lines. I like to use `.present?` to prevent that, but maybe that's not the case.

wwahammy: I suspect that wouldn't happen BUT there's no harm in adding `.present?` so I will!

PR #671, opened 2021-08-26
Add /api/nonprofits/:nonprofit_id/events/:id

wwahammy:
Closes #668﻿

PR #672, opened 2021-08-27
feature/api/ticket levels

### LINKED ISSUE #660: [FEATURE] Add api/nonprofits/:nonprofit_id/events/:event_id/ticket_levels and ./:id
Depends on #668

wwahammy:
﻿Closes #660

Depends on #671

### DIFF
--- a/app/models/ticket_level.rb
+++ b/app/models/ticket_level.rb
@@ -48,6 +48,10 @@ class TicketLevel < ApplicationRecord
     self.amount = 0 if amount.nil?
   end
 
+  def amount_as_money
+    Amount.new(amount||0, nonprofit.currency)
+  end
+
   # TODO replace with discard gem
   def discard!
     run_callbacks(:discard) do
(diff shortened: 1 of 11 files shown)


app/models/ticket_level.rb
```diff
@@ -48,6 +48,10 @@ def event_discounts
     self.amount = 0 if amount.nil?
   end
 
+  def amount_as_money
+    Amount.new(amount||0, nonprofit.currency)
+  end
```
clarissalimab, line 53: I think it's a good idea to organize the money like this. In the future we could move this to a concern.

clarissalimab (approved):
:rocket:, :speech_balloon::

app/models/ticket_level.rb
wwahammy: Agreed! Eventually, I think we should use the Money gem which will handle some of this stuff.

PR #673, opened 2021-08-31
Set up dependabot for Ruby dependencies

wwahammy:


clarissalimab (approved):
That's helpful! I think it's worth mentioning that one experience I had with dependabot was that  those daily checks were too much for our amount of gems we had and we didn't have enough time to review them properly every day.

wwahammy:
@clarissalimab maybe weekly would be a better idea?

PR #739, opened 2021-09-13
Remove unneeded yarn install in ruby.yml

wwahammy:


wwahammy:
Oops, this is actually wrong.

PR #749, opened 2021-09-17
Move rubocop into its own workflow 

wwahammy:
I moved rubocop out into a separate workflow job and integrated with the code scanning functionality of Github. Not sure this is the best design but I'm submitting it here for review for us to decide.

What do you think @clarissalimab ?

### DIFF
--- a/.github/workflows/placeholders.yml
+++ b/.github/workflows/placeholders.yml
@@ -45,3 +45,12 @@ jobs:
         ruby: [2.7.4]
     steps:
       - run: 'echo "Placeholder for main_build"'
+  rubocop:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


clarissalimab (commented):
I think it's better! The tests are going to run even if rubocop fails, it looks better to me. But why it looks like it didn't run and says "1 analysis not found"?

wwahammy:
> I think it's better! The tests are going to run even if rubocop fails, it looks better to me. But why it looks like it didn't run and says "1 analysis not found"?

So the code scanning feature of Github seems to try to compare two different builds to see if there are changes in the code scan result. It says an analysis is missing because there isn't one for the previous build because we hadn't used this feature until now. So I assume once we merge and then have a new PR, it won't say that anymore.

clarissalimab (approved):
:rocket:

PR #763, opened 2021-09-22
Add supporter for paging to /api/nonprofit/:nonprofit_id/supporters

wwahammy:
We use [kaminari](https://github.com/kaminari/kaminari) for paging as it's well supported.

We'll also merge each major api piece in order to break up the feature better.

Here we're doing paging for: `/api/nonprofit/:nonprofit_id/supporters`


clarissalimab (approved):
:rocket:

Issue #40, opened 2018-05-07
Event Header/Hide Header Text Bug


wendylbolm:
Expectations: You should always be able to modify customizations to pages that you've set. 

Reality: If you go to an Event and add a header image, you'll see an option in the event settings to hide the header text.

If you delete that image, the option to hide the header text (a checkbox) will disappear. However, if you have deigned to hide the header text, the option to show the header text will be gone, so you won't be able to add the header text back unless you upload another photo and go into the settings before you delete the photo.

nicaloribeiro:
Hi @wendylbolm ,  @iagotheophilo  and I would like to try to resolve this issue. However, we are having some trouble getting to the bug. In our view, it is necessary to create an event to have access to customizations, but it is necessary to confirm the email in order to create an event or campaign. The problem is that we couldn't confirm the email, we didn't receive the verification email.
Is this bug still present? If so, what would be the way to try to reproduce it?
We were also looking at issue #96, but we had the same email confirmation problem.
Thank you in advance for your attention.

marcelog5:
Hi @wendylbolm , @nicaloribeiro and I when try to create a event we are getting this error:

![Sem título](https://user-images.githubusercontent.com/48961541/134079986-7955235c-ca2b-4969-a3a5-acd60078c749.png)

and the console says:
```
app/controllers/events_controller.rb:40:in `block in create'
method=POST path=/nonprofits/2/events format=*/* controller=EventsController action=create status=500 error='SystemStackError: stack level too deep' duration=40.11 view=0.00 db=4.51 time=627.70 exception=["SystemStackError", "stack level too deep"] exception_object=stack level too deep

SystemStackError (stack level too deep):

app/controllers/application_controller.rb:59:in `render_json'
app/controllers/events_controller.rb:34:in `create'
```

this happens after filling in the fields and clicking on "Preview event"

Do you know what wee are doing wrong to get this error.

wwahammy:
Hi there, @marcelog5, I can check into this right now!

wwahammy:
@marcelog5 I've proposed a fix on https://github.com/houdiniproject/houdini/issues/40#issuecomment-923350246 to address this.

There is still a bug related to loading the ticket levels on the event page that I need to address.

wwahammy:
@marcelog5 I think I've fixed loading the ticket levels. There's some quirkiness here and there but it's workable.

wwahammy:
[...]

PR #765, opened 2021-09-22
Fix bug preventing creation of events

wwahammy:
Fixes the problem highlighted in #40 (﻿https://github.com/houdiniproject/houdini/issues/40#issuecomment-923350246).

There's still a bug related to ticket_levels not loading on the event page but this fixes the immediate issue.


clarissalimab (approved):
:rocket:

PR #775, opened 2021-09-29
Remove no longer supported docker instance

wwahammy:
We don't really use docker anymore and it's not really supported﻿. I don't think it's helpful to have in the repo in that case so I think we should remove it.

I'll keep this open for a few days for comments.


README.md
```diff
@@ -56,8 +56,6 @@ Check on [contribution_guide_usability_testing.md](docs/contribution_guide_usabi
 
 #### Tips for specific circumstances
 
-* Docker: Docker was previously used for development of Houdini. 
-See [docker.md](docs/docker.md) for more info.
 * Mac: Mac dev setup may require some unique configuration. 
 See [mac_getting_started.md](docs/mac_getting_started.md) for more info.
```
clarissalimab, line 60: I think we should remove this entire session or update it, this `mac_getting_started.md` does not exist anymore.

clarissalimab (approved):
I agree with removing it. I just left a suggestion:

wwahammy:
As there are no comments, I will merge this.

PR #780, opened 2021-09-30
Fix bug on Campaign updating

### LINKED ISSUE #778: [BUG] Updating campaign causes server to crash
Trying to update a campaign through the UI results in a "Stack level too deep" loop causing the server to crash.

## Description

Trying to update a campaign results in a "Stack level too deep" loop causing the server to crash. 
The campaign object does get updated, however the server must be restarted to recover from the crash. 

## Steps to Reproduce

1.
2.
3.
4.

## Your Environment

* Ruby version:
* Server type and version:
* Operating System and version:
* Link to your project:

wwahammy:
Fixes #778﻿


clarissalimab (approved):
Looks good! :rocket:

PR #792, opened 2021-10-05
Only use eslint-jest/style and ./recommended rules

wwahammy:
Currently, we're use all eslint-plugin-jest rules and some of them aren't that useful. I tried to move the few that are useful into the config.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,12 +11,15 @@ const tsSpecBase = {
 	extends: [
 		'eslint:recommended',
 		'plugin:@typescript-eslint/recommended',
-		'plugin:jest/all',
+		'plugin:jest/recommended',
+		'plugin:jest/style',
 		'plugin:react-hooks/recommended',
 	],
 	rules: {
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -11,12 +11,15 @@ const tsSpecBase = {
...
-		'plugin:jest/all',
+		'plugin:jest/recommended',
+		'plugin:jest/style',
 		'plugin:react-hooks/recommended',
 	],
 	rules: {
 		"jest/lowercase-name": ["error", { "ignore": ["describe"] }],
 		"react-hooks/exhaustive-deps": 'error',
+		"jest/no-duplicate-hooks": 'error',
+		"jest/prefer-hooks-on-top": 'warning',
```
clarissalimab, line 22:
```suggestion
		"jest/prefer-hooks-on-top": 'warn',
```

clarissalimab (approved):
Nice!

.eslintrc.js
wwahammy: Thank you for catching that!

PR #793, opened 2021-10-06
Replace some methods names

### LINKED ISSUE #783: [FEATURE] Meaningful methods names
**Is your feature request related to a problem? Please describe.**
I came across some methods that I didn't understand their purpose until I read their implementation

**Describe the solution you'd like**
I thought of some alternatives and I would like to know what you think:
[`simple_date`](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L31) -> `format_date_to_mdY`
[`readable_date`](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L45) -> `format_date_with_month_name`

And some others. If you think that it can be implemented, I can work on that and any other problem related to this.

heronsousa:
Update some methods to more meaningful names to be more clear their purpose.

Closes #783


heronsousa:
> 
> 
> It looks pretty good but one tweak would be to put a comment before each of the methods so we can understand exactly what they do. Otherwise, even with the names, we might be confused.

Sure, I understand your point. But trying to follow what clean code propose, the ideia is to have methods names that tells why it exists, what it does, and how it is used, doesn't needing comments. If the the names are not clear enough, I can change than.

wwahammy:
> @wwahammy sure, I understand your point. But trying to follow what clean code propose, the ideia is to have methods names that tells why it exists, what it does, and how it is used, doesn't needing comments. If the the names are not clear enough, I can change than.

I definitely agree that the names should be able to tell you what they do. In my experience, there are few situations where documentation isn't needed in addition to clear naming. For example, let's say someone jumps into the file and doesn't have the same experiences that make what you or I think are self-descriptive names. To them, the name doesn't make sense and then they don't have any idea what the code does without looking at it.

Additionally, I think these functions have outputs which are really too difficult to describe solely with the function name.

heronsousa:
> I definitely agree that the names should be able to tell you what they do. In my experience, there are few situations where documentation isn't needed in addition to clear naming. For example, let's say someone jumps into the file and doesn't have the same experiences that make what you or I think are self-descriptive names. To them, the name doesn't make sense and then they don't have any idea what the code does without looking at it.
> 
> Additionally, I think these functions have outputs which are really too difficult to describe solely with the function name.

Now I understand what you mean, I needed to see the documentation again for some date simbols hahaha. I will add the comments and an example for each one.

wwahammy:
Hey there  @heronsousa, any update on this change? We'd really love to get it merged!

WeillerFernandes:
Hello @wwahammy, @GuilhermeBraz and I have been working on this issue for the last week and we opened another Pull Request (https://github.com/houdiniproject/houdini/pull/1002) related to it, which would make this Pull Request deprecated.

wwahammy:
This was handled by #1002.

PR #836, opened 2021-10-30
Feature build traceroute

marcelog5:
please don't merge this pull request!!!

This pull request have the objective to test the job traceroute in the build process in the ruby workflow


.github/workflows/ruby.yml
```diff
@@ -1,6 +1,6 @@
 # License: LGPL-3.0-or-later
 name: Ruby test and build
-on: 
+on:
```
wwahammy, line 3: Make sure to remove the reformatting here. There's nothing really wrong with the end result but it clutters the PR

.github/workflows/ruby.yml
```diff
@@ -86,5 +86,22 @@ jobs:
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Run traceroute to verify unused routs
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+      - name: Traceroute run
+        run: rake traceroute
```
wwahammy, line 107: Based on the error provided by the workflow action, I think this should be `bundle exec rake traceroute`.

wwahammy (changes requested):
Great start! Just a few things left to do.

PR #845, opened 2021-11-02
Remove billing subscriptions

wwahammy:
Billing subscriptions of individual nonprofits are not a core feature of Houdini and they make the project more complex by having it in the core part of Houdini. I support adding a billing subscription plugin later but I'm not a fan of having it in core since it really only has use for companies.

To that end, I've removed﻿ the billing subscription features. I don't think it will cause a lot of problems for any users and it will simplify setup for the average user.

In case anyone was wondering, I've kept the BillingSubscription class because it's really just a join table between Nonprofit and BillingPlan. BillingPlan should get some cleanup as well but that's a separate issue.


wwahammy:
Assuming this is approved and there are no complaints, I will merge this on November 8, midday US Central Time.

clarissalimab (approved):
Looks good!

PR #850, opened 2021-11-03
Rename the id param in APIs to transaction_id and create a `current_transaction` concern

wwahammy:
Currently, we use the id param in the API for the last item in a route. For example, if you wanted a transaction via the API, it would be `﻿/api/nonprofits/:nonprofit_id/transactions/:id`. This is normally fine but it gets awkward when you have a nested resource. Then it becomes: `/api/nonprofits/:nonprofit_id/transactions/:transaction_id/subtransactions/:id`. This makes it difficult have a single method like `current_transaction` to get the current transaction because the param being checked varies.

This PR changes changes the routes so use  `/api/nonprofits/:nonprofit/transactions/:transaction_id`. This also allows us to create a concern to get current transaction via `current_transaction`.

### DIFF
--- a/app/controllers/api/transactions_controller.rb
+++ b/app/controllers/api/transactions_controller.rb
@@ -5,7 +5,7 @@
 
 # A controller for interacting with a nonprofit's supporters
 class Api::TransactionsController < Api::ApiController
-	include Controllers::Nonprofit::Current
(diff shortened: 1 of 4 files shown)


clarissalimab (approved):
Good job!

PR #859, opened 2021-11-07
Upgrade to Typescript 4.4.4

wwahammy:


wwahammy:
There's a bug here so I'm going to close and investigate

PR #878, opened 2021-11-16
Correct partial specs so they have the same namespaced prefix in view specs versus request specs

wwahammy:
The most natural way to include a partial in jbuilder is as follows:

```ruby 
﻿json.partial! @user, as: :user
```

In this case, Rails looks at the type of `@user` which has a class name of "User" and then looks into the views directory for a partial ending 'users/_user.json.jbuilder".

The mechanism rails uses to decide the prefix when handling request for the a controller is to take the nearest routing namespace (or none if you're at the root) and then prefix the partial for searching (it's more complex than that but for our purposes that's good enough). So, if the the controller is `api/users_controller.rb` and the partial above is the view for the `current` action (in `app/views/api/users/current.json.jbuilder` then it would be looking for a partial in the `app/views/api/users/_user.json.jbuilder` (it would also be looking inside the views folder of all of the engines.

This mechanism works fine in controller specs but, for some reason, fails badly in view specs. In that case, the prefix is filled with three different values: ["", "<some path in the action_view module>", "/api/users"]. For unclear reasons, this leads to view specs failing.

The fix I used was to go to all of the specs where we're going to be calculating the proper view specs is to remove the first two prefixes with the following line before the `render` happens:

```ruby
view.lookup_context.prefixes = view.lookup_context.prefixes.drop(2)
```

Is this hacky? Incredibly. I don't know why the lookup_context is messed up in the view specs but this fixed it for our use-case.

Given that fix, I then moved all of the `show.json.jbuilder` views to use the autosensing partials feature.


clarissalimab (approved):
Sorry, I looked at it and forgot to approve a couple of days ago. Approved now! :rocket:

PR #879, opened 2021-11-16
Make new role (user) modal roles/_new_modal.html.erb translatable using i18n

clarissalimab:


app/views/roles/_new_modal.html.erb
```diff
@@ -13,25 +13,29 @@
...
+					<%= t('roles.new.important') %>
+					<br>
+					<%= t('roles.new.double_check_this_email_address') %>
+				</strong></p>
+				<input name='role[email]' type='email' placeholder="<%= t('roles.new.email') %>" id='role-email-field' required><br>
+				<input name='role[email_confirmation]' type='email' placeholder="<%= t('roles.new.confirm_email') %>" id='role-email-field' required parsley-equalto='#role-email-field'>
 
 				<p class='error'></p>
-				<%= render 'common/modal_submit', loading: 'Creating...', text: 'Create' %>
+				<%= render 'common/modal_submit', loading: t('roles.new.creating'), text: t('roles.new.create') %>
```
wwahammy, line 38: One minor tweak: "Create" is such a common UI text element that it might make sense to add it to ui.en.yml and refer to it from there. 

Thoughts?

app/views/roles/_new_modal.html.erb
```diff
@@ -13,25 +13,29 @@
...
 				</div>
 
-				<p><strong>Important!<br/> Double check this email address</strong></p>
-				<input name='role[email]' type='email' placeholder='Email' id='role-email-field' required><br>
-				<input name='role[email_confirmation]' type='email' placeholder='Confirm email' id='role-email-field' required parsley-equalto='#role-email-field'>
+				<p><strong>
+					<%= t('roles.new.important') %>
+					<br>
+					<%= t('roles.new.double_check_this_email_address') %>
+				</strong></p>
```
wwahammy, line 33: Is there any language that we can foresee where the order of these two UI elements would be reversed?

To be clear, we don't need to evaluate every language that exists, I'd just like to avoid a situation where we have 10 languages and then we realize we have to change all of them because a new language we add expects it in the opposite order.

wwahammy (commented):
👍

I had two comments. Once we get those two clarified, it's ready to merge.

app/views/roles/_new_modal.html.erb
clarissalimab: Makes sense!

app/views/roles/_new_modal.html.erb
clarissalimab: I think it would be the case if it was a single sentence instead of two, but I don't know if there's any language that would make that different. It behaves like two different HTML elements, I think the person chose to use `<br>` instead of two `<p>` elements because it was easier than handling the spacing.

wwahammy: Hmmm, I think we should go with two elements and then work through the problem later if it exists.

app/views/roles/_new_modal.html.erb
clarissalimab: I think it applies "Creating..." too

app/views/roles/_new_modal.html.erb
```diff
@@ -13,25 +13,29 @@
...
+				<input name='role[email]' type='email' placeholder="<%= t('roles.new.email') %>" id='role-email-field' required><br>
```
wwahammy, line 34: Just a reminder, it's best to not inject Ruby values into an attribute this way. It works but it's easy to mess up. You really don't need to fix it here but, in this case, you could have used [text_field_tag](https://api.rubyonrails.org/v6.1.4/classes/ActionView/Helpers/FormTagHelper.html#method-i-text_field_tag). It's less awkward because you don't have to worry about quoting around your erb tags.

wwahammy (approved):
Looks good! Go ahead and merge.

PR #880, opened 2021-11-17
Translates roles.new strings to Brazilian Portuguese (pt_BR)

clarissalimab:


wwahammy (commented):
Those are indeed words. I assume they are the correct words 😂

config/locales/roles.new.pt_BR.yml
```diff
@@ -0,0 +1,44 @@
...
+#   'true': 'foo'
+#
+# To learn more, please read the Rails Internationalization guide
+# available at https://guides.rubyonrails.org/i18n.html.
+
+pt_BR:
+  roles:
+    new:
+      invite_a_new_user: Convide um Novo Usuário
+      nonprofit_admin: Administrador da Organização
```
clarissalimab, line 37: The literal translation of "nonprofit" does not make sense in Brazilian Portuguese, so I translated to "Organização", which is equivalent to "Organization". What do you think?

config/locales/roles.new.pt_BR.yml
```diff
@@ -0,0 +1,44 @@
...
+      full_permissions_including_bank_payouts: "Todas as permissões, incluindo pagamentos bancários"
```
clarissalimab, line 38: The double quotes were automatically added by Transifex.

config/locales/roles.new.pt_BR.yml
wwahammy: In some places they're called non-governmental organizations so is there a word for that? But you really know best so whatever you decide is good in my book 🙂

wwahammy: One thing to consider is making sure whatever phrasing you use is consistent. It probably should be documented somewhere that this is the word to use.

clarissalimab: For non-governmental organizations it would be ONGs (initials for Organização Não Governamental), but it is technically wrong to call that - at least in Brazil - if they are not formally instituted like that, that's why I thought about "Organizations", which would be a more broad name.

> It probably should be documented somewhere that this is the word to use.

We could have a glossary for that.

wwahammy: I think Organização is good then!

wwahammy:
@clarissalimab where is this PR on merging?

clarissalimab:
@wwahammy I think I won't merge it for now while I'm testing the automated process

PR #889, opened 2021-11-25
Use http_accept_language gem to extract locale from request headers

clarissalimab:
This is based on [this guide](https://guides.rubyonrails.org/i18n.html#choosing-an-implied-locale).

### DIFF
--- a/app/controllers/concerns/controllers/locale.rb
+++ b/app/controllers/concerns/controllers/locale.rb
@@ -2,22 +2,35 @@
 
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
-# rubocop:disable Style/ConditionalAssignment
 module Controllers::Locale
 	extend ActiveSupport::Concern
 
 	included do
-		before_action :set_locale
+		around_action :switch_locale
 
 		private
 
-		def set_locale
-			if params[:locale] && Houdini.intl.available_locales.include?(params[:locale])
-				I18n.locale = params[:locale]
-			else
-				I18n.locale = Houdini.intl.language
-			end
+		# We decide on the locale based on the request headers, extracted by 'http_accept_language' gem,
+		# unless the request comes with a locale param, then we override it.
+		def switch_locale(&action)
+			locale = if available_locales.include?(params[:locale])
+													params[:locale].tr('-', '_')
+												else
+													extract_locale_from_accept_language_header
+												end
+			logger.debug "* Locale set to '#{locale}'"
+			I18n.with_locale(locale, &action)
+		end
+
+		def extract_locale_from_accept_language_header
+			require 'http_accept_language' unless defined? HttpAcceptLanguage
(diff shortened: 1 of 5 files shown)


app/controllers/concerns/controllers/locale.rb
```diff
@@ -2,22 +2,27 @@
...
+													params[:locale]
+												else
+													http_accept_language.compatible_language_from(Houdini.intl.available_locales)
+												end
+			locale = adequate_locale_to_transifex_syntax(locale)
+			logger.debug "* Locale set to '#{locale}'"
+			I18n.with_locale(locale, &action)
+		end
+
+		def adequate_locale_to_transifex_syntax(locale)
```
clarissalimab, line 24: Although the default keys from locales that include variants such as Brazilian Portuguese is to use a dash (`-`) to determine the variant (as in `pt-BR`), Transifex uses underscore.

wwahammy: Is it possible to change the underscores from Transifex to dashes before merging? I think it would be clearer if it worked that way.

app/controllers/concerns/controllers/locale.rb
```diff
@@ -2,22 +2,27 @@
...
 
 		private
 
-		def set_locale
-			if params[:locale] && Houdini.intl.available_locales.include?(params[:locale])
-				I18n.locale = params[:locale]
-			else
-				I18n.locale = Houdini.intl.language
-			end
+		def switch_locale(&action)
```
wwahammy, line 13: It'd be good to document this algorithm as a comment on this method.

wwahammy (changes requested):
I really like this! In particular, I like the way you did this so that passing the param overrides the header.

app/controllers/concerns/controllers/locale.rb
clarissalimab: I couldn't find this option anywhere :(

wwahammy:
So I think `http_accept_language` is probably simple enough that we can get away with it BUT it hasn't been updated since 2017. That worries me a bit.

clarissalimab:
@wwahammy Hmm, I didn't notice. It was a gem recommended for that in the most recent Rails guide version. They also mention using [this](https://github.com/rack/rack-contrib/blob/master/lib/rack/contrib/locale.rb) as an alternative. What do you think?

wwahammy:
@clarissalimab given the number of people using the gem (https://rubygems.org/gems/http_accept_language/versions/2.1.1) I think we should use it.

wwahammy:
It might be useful to look at how discourse does it and take their code: https://github.com/discourse/discourse/blob/1472e47aae5bfdfb6fd9abfe89beb186c751f514/lib/http_language_parser.rb

clarissalimab:
@wwahammy [that line](https://github.com/discourse/discourse/blob/1472e47aae5bfdfb6fd9abfe89beb186c751f514/lib/http_language_parser.rb#L5) explains why Transifex uses underscore instead of hyphen.

wwahammy (approved):
Looks good to me!

PR #910, opened 2021-12-09
Correct some things in useCurrentUser/Auth

### LINKED ISSUE #904: [BUG] Inconsistent state when running react tests
## Description
Occasionally, on builds, we have  a jest test that fails. One example is [here](https://github.com/houdiniproject/houdini/runs/4269682856?check_suite_focus=true). I suspect this is due to the `act` warnings we consistently receive. 

## Expected Behavior
Tests either pass reliably.

## Actual Behavior
Intermittent test failures.

## Possible Fix
Fix act bugs using the description at https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning. I suspect this also has something to do with the way formik works.

wwahammy:
Closes #904 (I think)

- useCurrentUser.revalidate uses the proper mutate arguments
- Migrate useCurrentUserAuth to useReducer and correct a promise bug

### DIFF
--- a/app/javascript/hooks/useCurrentUser.ts
+++ b/app/javascript/hooks/useCurrentUser.ts
@@ -78,7 +78,7 @@ function useCurrentUser<TReturnType extends UseCurrentUserReturnType = UseCurren
 	const currentUser = error?.status === NotLoggedInStatus ? null : data;
 
 	async function revalidate() {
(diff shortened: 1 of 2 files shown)


clarissalimab (approved):
Nice!

PR #913, opened 2021-12-09
Migrate from Formik to react-hook-form

wwahammy:
Formik has some bugs around promises that make tests unreliable. As recommended by @tttp, I've attempted to migrate from Formik to react-hook-form in this PR. I think this is far easier and strongly believe we should use this going forward.

I'll obviously squash these commits going forward before merging if it's approved.

Depends on #912


clarissalimab (approved):
Nice!

PR #922, opened 2021-12-15
Update to Ruby 2.7.5

wwahammy:
Ruby 2.7 has a new patch version, 2.7.5, for fixing a few security issues. This PR updates to that.

Additionally, I've removed the unneeded Ruby version listed in the Gemfile since I'm here. That version prevents people from easily testing on other Ruby versions which may be useful for some folks.

Depends on #926


wwahammy:
@clarissalimab when this is otherwise ready to merge, we'll have to switch the required checks to use 2.7.5 so this is mergeable.

clarissalimab (approved):
:rocket:!

PR #923, opened 2021-12-15
Update to React 17

wwahammy:
Depends on #919

Depends on #925

Depends on #926 

Depends on #928

This updates React to React 17. Wheee.


wwahammy:
I think I'm ready for a review here too @clarissalimab!

clarissalimab (approved):

PR #925, opened 2021-12-16
Simplify the useCurrentUserAuth hook

wwahammy:
By using useAsyncFn, we simplify our hook and standardize.﻿


clarissalimab (approved):
So cleaner!

PR #934, opened 2021-12-20
Work in progress on AmountStep and InfoStep

clarissalimab:
This is a WIP for the migration of the Donate Wizard to React. It includes:

* Storybook support with a few different scenarios;
* Wizard styling;
* AmountStep including custom amounts, or inserting a different amount, selecting recurring donation;
* InfoStep including the supporter fields component, different payment methods and other details.

This is how it looks now:
donate-wizard

The style on InfoStep looks off because we're still not using the select input to choose the country and the state and the payment buttons need a review to check what's missing on the styles.

What's next:
* Post the supporter when choosing a payment method;
* Add designation on the AmountStep;
* Missing components on the InfoStep;
  * CustomFields;
  * DedicationLink;
  * AnonymousCheckbox;
  * DedicationForm;
* Fix the styles on the InfoStep;
* Load countries and cities information on the InfoStep;
* PaymentStep;
* FollowupStep.

### DIFF
--- a/app/javascript/components/legacy/nonprofits/donate/InfoStep.tsx
+++ b/app/javascript/components/legacy/nonprofits/donate/InfoStep.tsx
@@ -1,7 +1,9 @@
 // License: LGPL-3.0-or-later
-import React from 'react';
+import { Money } from '../../../../common/money';
+import React, { useContext } from 'react';
 import { useIntl } from "../../../intl";
-import { SupporterType, RequiredFieldsType } from './wizard';
+import { SupporterType, RequiredFieldsType, DonationWizardContext } from './wizard';
+import { WizardContext } from '../../_dependencies/ff-core/wizard';
 
 // const h = require('snabbdom/h')
 // const R = require('ramda')
@@ -12,7 +14,6 @@ import { SupporterType, RequiredFieldsType } from './wizard';
 // const dedicationForm = require('./dedication-form')
 // const serialize = require('form-serialize')
 // const request = require('../../common/request')
-// const format = require('../../common/format')
 
 // const sepaTab = 'sepa'
 // const cardTab = 'credit_card'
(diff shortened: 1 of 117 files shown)


app/javascript/components/legacy/nonprofits/donate/InfoStep.tsx
```diff
@@ -0,0 +1,379 @@
...
+					type: 'setSelectedPayment',
+					selectedPayment: values.selectedPayment,
+				});
+				props.stateDispatch({
+					type: 'setSupporter',
+					supporter: values.supporter,
+					address: values.address,
+					next: stepManagerContext.next,
+				});
+			}} initialValues={{ supporter: props.supporter, address: props.address } as FormikFormValues}>
```
wwahammy, line 60: Important note: in React you should always set the form field default values; if the value is null or undefined, then it should be a blank string. React has problems tracking form field values when they're originally set to null or undefined (I don't totally understand it but every bit of advice has told me that).

If the service or whatever is getting the result of the submit expects that those values will be null or undefined, then you should convert empty form fields back to null or undefined on submit.

app/javascript/components/legacy/nonprofits/donate/InfoStep.tsx
```diff
@@ -0,0 +1,379 @@
...
+						name="email"
+						title={emailTitle}
+						required={props.required.email}
+						value={supporter?.email}
+						placeholder={emailTitle}
+						onChange={
+							(e) => {
+								setSupporterFields('email', e.target.value);
+							}
+						} />
```
wwahammy, line 137: Handling input this way avoids using Formik entirely (or in the near future, react-hook-form) which isn't good because those tools give us things that help us. While it's likely not useful to investigate how Formik works at this point since I hope to have it switched over to react-hook-form today, the same principles apply there.

app/javascript/components/legacy/nonprofits/donate/amount-step.tsx
```diff
@@ -1,52 +1,67 @@
...
+import { WizardContext } from '../../_dependencies/ff-core/wizard';
+import BigNumber from 'bignumber.js';
+
 interface AmountStepProps {
-	amount: Money|null;
+	amount: Money | null;
 	amountOptions: Money[];
+	stateDispatch: (action: ActionType) => void;
+	currencySymbol: string;
+	singleAmount: string | null;
```
wwahammy, line 16: This should be `Money | null`. Everything else here uses Money so let's be consistent.

app/javascript/components/legacy/nonprofits/donate/InfoStep.tsx
clarissalimab: You mean it should be something like this?

```suggestion
					<input
						type="email"
						name="email"
						title={emailTitle}
						required={props.required.email}
						value={values.supporter.email}
						placeholder={emailTitle}
						onChange={
							(e) => {
								setFieldValue('supporter.email', e.target.value); // not sure if this is going to work
							}
						} />
```

wwahammy: We're talking about this right now :) Look at using the Field Component in Formik. :smile:

PR #918, opened 2021-12-10
New Crowdin updates

wwahammy:
This is the Acholi pseudo-language. I think I have this the right way. What do you think @clarissalimab?

PR #948, opened 2021-12-31
Document translations process

### LINKED ISSUE #840: [BUG] Document translation process
Currently, our translation process uses a [Transfex project](https://www.transifex.com/houdini-project/houdini/dashboard/) for managing it. I think that, in principle, is perfectly fine. That said, we basically don't have a translation process or formal guidelines on how translations should work, where sources should be added, etc.

I generally think it should follow the following design:

* Strings get added to the English translations in `config/locales`
  * If a new locale file is added, it needs to be added to Transifex as a new resource
* When new English strings are added, they should be synced to Transifex.
* When new translations are created, a PR should be created with the change to merge in the change.

Some open questions:
* When is a string added? Once merged into main?
* What's the mechanism for updating those strings? If manual, what's the steps to make sure nothing falls through the cracks?
* How do we add new people to the translation process?
* How are PRs created for updates to translations? If manual, what's the steps to make sure nothing falls through the cracks?
* How do people decide when to create a new locale file? Generally I think it should be a different locale file for each significant UI component.

clarissalimab:
Closes #840.

Depends on https://github.com/houdiniproject/houdini/pull/918.

Depends on #1018

### DIFF
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -1,57 +1,73 @@
+<!-- markdownlint-disable MD029 -->
+
 # Translating Houdini
 
-We use [Crowdin](https://crowdin.com/project/houdiniproject) to work on our translations. Not all Houdini's pages and components are translatable, we are working on making Houdini entirely translatable, and you can help us with that too!
+We use [Crowdin](https://crowdin.com/project/houdiniproject) to work on our
+translations. Not all Houdini's pages and components are translatable, we are
+working on making Houdini entirely translatable, and you can help us with that
+too!
 
 ## Getting started
 
-If you want to help Houdini with translations, you can do it by using our [Crowdin](https://crowdin.com/project/houdiniproject) to add your translations (see [Translating strings using Crowdin](#translating-strings-using-crowdin)).
+If you want to help Houdini with translations, you can do it by using our
+[Crowdin](https://crowdin.com/project/houdiniproject) to add your translations
+(see [Translating strings using Crowdin](#translating-strings-using-crowdin)).
 
-If a page, component, or string is not available on Crowdin to be translated yet, you're going to need to follow the [Making a component translatable](#making-a-component-translatable) guide to creating the source translation file if needed or add the keys to a source translation file and add support to i18n to the component before adding translations to another language.
+If a page, component, or string is not available on Crowdin to be translated
+yet, you're going to need to follow the [Making a component
+translatable](#making-a-component-translatable) guide to creating the source
+translation file if needed or add the keys to a source translation file and add
+support to i18n to the component before adding translations to another language.
 
-If there's a translation that you'd like to change, take a look at the [Modifying translations](#modifying-translations) topic.
+If there's a translation that you'd like to change, take a look at the
+[Modifying translations](#modifying-translations) topic.
 
 ## Translating strings using Crowdin
 
-> To translate strings, they need to be available in Crowdin. If they are not yet, you can help us by making them translatable by following [this guide](#making-a-component-translatable) and then translating them to the language that you want.
+> To translate strings, they need to be available in Crowdin. If they are not
+> yet, you can help us by making them translatable by following [this
+> guide](#making-a-component-translatable) and then translating them to the
+> language that you want.
 
-1. From [Crowdin](https://crowdin.com/project/houdiniproject) main page, select the language to which you want to translate.
+1. From [Crowdin](https://crowdin.com/project/houdiniproject) main page, select
+   the language to which you want to translate.
 
-> Note: If the language still doesn't exist on Crowdin, you can request that a language gets added by [filing an issue](https://github.com/houdiniproject/houdini/issues/new?assignees=&labels=enhancement&template=language_request.md&title=%5BFEATURE%5D).
+> Note: If the language still doesn't exist on Crowdin, you can request that a
+> language gets added by [filing an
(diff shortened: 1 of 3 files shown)


docs/translations.md
```diff
@@ -0,0 +1,103 @@
...
+
+Examples:
+
+* On .erb files, you should replace the "Full permissions, including bank payouts" text for `t('roles.new.full_permissions_including_bank_payouts')`;
+* For `.tsx` files, you should import the `useIntl` hook from the `intl` component and import `formatMessage` from it. You should then create a constant with `formatMessage({ id: 'roles.new.full_permissions_including_bank_payouts' })` and replace the text by the name of the constant you created - you can look at how it's done on our [SignInComponent.tsx](https://github.com/houdiniproject/houdini/blob/a31e755f1bdf21c5c894018fe8ec3b26fcf6c896/app/javascript/components/users/SignInComponent.tsx#L77);
+* For `.js` or `.es6` files, replace the text with `I18n.t('roles.new.full_permissions_including_bank_payouts')`.
+
+7. Commit the changes and open a pull request to Houdini/main.
+
+8. Once the pull request is merged, after up to 1 hour, Crowdin is going to sync the new file and make it available for translations. Once the file is synced, you can start translating from Crowdin.
```
clarissalimab, line 95: I considered as we would be using our current file organization instead of having folders for each language. I saw some other Rails projects organize like that, so I think we could keep it. [Discourse](https://github.com/discourse/discourse/tree/main/config/locales) is an example, and they use their Crowdin instance for that.

Here's the text if we change our minds and decide to change the organization:

```suggestion
We can also have translations for strings that are shared in multiple components. For example, [config/locales/ui.en.yml](config/locales/en-EN/ui.yml) holds UI strings that can be shared between multiple components, such as common button labels or links.

2. If **there are** translation files where the strings you want to make translatable can be placed (given the context and the organization of the files), you can add the keys and original text to this file (see the example below). **After adding the new keys and strings, you can go to step 4.**

Example:

```yml
en:
  roles:
    new:
      invite_a_new_user: Invite a New User
      nonprofit_admin: Nonprofit Admin
      ...
\```

3. If **there aren't** any translation files for the strings that you want to make translatable, then you need to create a new source file. The name of the translation file should be the same as the component name, but with `.en` in the and, followed by the `.yml` extension. For example, [app/views/roles/_new_modal.html.erb](app/views/roles/_new_modal.html.erb)'s translation file is named as `roles.new.en.yml`.

Example:

```yml
en:
  roles:
    new:
      invite_a_new_user: Invite a New User
      nonprofit_admin: Nonprofit Admin
      ...
\```

4. Find every string that is going to be translated, create a key for it and place it into your translation file.

It's important to give it a well detailed name, so that when someone is reading the code on the component they automatically understand what that string is without having to look at the translation file.

For example, the text `Full permissions, including bank payouts` on [roles.new.yml](config/locales/en-EN/roles.new.yml), as big as it is, has the key `full_permissions_including_bank_payouts`.

5. Replace the strings from the components with their corresponding keys.

Examples:

* On .erb files, you should replace the "Full permissions, including bank payouts" text for `t('roles.new.full_permissions_including_bank_payouts')`;
* For `.tsx` files, you should import the `useIntl` hook from the `intl` component and import `formatMessage` from it. You should then create a constant with `formatMessage({ id: 'roles.new.full_permissions_including_bank_payouts' })` and replace the text by the name of the constant you created - you can look at how it's done on our [SignInComponent.tsx](https://github.com/houdiniproject/houdini/blob/a31e755f1bdf21c5c894018fe8ec3b26fcf6c896/app/javascript/components/users/SignInComponent.tsx#L77);
* For `.js` or `.es6` files, replace the text with `I18n.t('roles.new.full_permissions_including_bank_payouts')`.

6. Commit the changes and open a pull request to Houdini/main.

7. Once the pull request is merged, after up to 1 hour, Crowdin is going to sync the new file and make it available for translations. Once the file is synced, you can start translating from Crowdin.
```

docs/translations.md
```diff
@@ -0,0 +1,105 @@
...
+4. Add the new file(s) you created to the [Crowdin config file](../crowdin.yml) indicating the path for the source language and the path for the translations.
+
+Example:
+
+```yml
+files:
+  - source: /**/locales/en.yml
+    translation: /**/locales/%locale%.yml
+  - source: /**/locales/roles.new.en.yml
+    translation: /**/locales/roles.new.%locale%.yml
```
wwahammy, line 76: I think it would be simpler if we didn't have to add a new source file to this `crowdin.yml` file. Ultimately, the pattern we want is:
```yml
 - source: /**/locales/%part_of_file_name_before_locale%.en.yml
    translation: /**/locales/%part_of_file_name_before_locale%.%locale%.yml
```
I don't know if that's possible but it would avoid someone forgetting to add every new source file.

clarissalimab: There's no way to do that, sadly :(

wwahammy: I think that would be possible if we reorganize the locales into directories.

wwahammy:
It would probably be good to document somewhere how we're supposed to handle more complex locales, like pt-BR, i.e. that we need to talk to Crowdin and what we need to ask them.

wwahammy:
One thing I thought of last night: make sure to mention that  html tags will be turned into HTML entities, i.e. `<` gets turned into `&lt;`. Generally you want to avoid using html in your translatable strings but in some cases, it's unavoidable, like if a single word is bolded in the middle of a sentence. In that case, the source key must end in `_html`. This is a requirement of Rails.

wwahammy:
Don't we have to merge in `crowdin-setup` first?

PR #958, opened 2022-01-04
Cleanup of lib folder

wwahammy:
* Removed a handful of empty or unused files
* Renamed the CCOrgError to HoudiniError as part of the de-CommitChange-ification of the codebase.

PR #960, opened 2022-01-04
Migrate to Zeitwerk

wwahammy:
Zeitwerk is the default autoloader in Rails 6 and mandatory in Rails 7. According to the Rails team, it simplifies and improves the performance of the Autoloader. This PR attempts to implement Zeitwerk.

Autoloading has always been a little funky in Houdini so this should help a bit.

Zeitwerk has specific needs for directories as it relates to a load path. In particular, you need one file per autoloaded constant and they must follow the pattern of `<camelcased_module_name>/<camelcased_autoloaded_filename>.rb`. For example, if you have a file with a class called: `Format::CsvHome`, it needs to be in the `format/csv_home.rb` file. Why is this important? Because inside our `lib` directory, we follow a directory format that doesn't follow this state. For example `UpdateDonations` is at `update/update_donations.rb`. To address this, I move all of the contents of the `lib` directory that don't relate to normal Rails `lib` folder contents into `app/legacy_lib` and flatten it. Is this the best way to do it? I'm not sure but I think overall it is.

While all the tests pass, I haven't actually run `bin/rails server` so someone should do that :)

I guess my main question for people is if my plan for moving files into `app/legacy_lib` makes sense to people or if there's a better way?

Depends on #958


clarissalimab (approved):
Nice work!

> While all the tests pass, I haven't actually run bin/rails server so someone should do that :)

I ran the server and it works properly!

> I guess my main question for people is if my plan for moving files into app/legacy_lib makes sense to people or if there's a better way?

`app/legacy_lib` makes sense to me.

wwahammy:
Awesome! I'll cleanup the rubocop results and merge!

PR #975, opened 2022-01-16
Added and ran fastener

wwahammy:
[fastener](https://github.com/DamirSvrtan/fasterer﻿) is a utility for finding performance improvements for a ruby codebase. I added it and ran it and these are the changes and improvements.

NOTE: There are a few improvements which I didn't apply because I thought they made the code more confusing.


clarissalimab (approved):
This is very cool!

Issue #798, opened 2021-10-07
[FEATURE] Cleaning unused code


lucasgandac:
**Is your feature request related to a problem? Please describe.**
On some files in houdini, there are some unused code that might be deleted to improve code cleanliness. 

**Describe the solution you'd like**
Remove old unused code.

**Additional context**
An example is  [index.tsx](https://github.com/houdiniproject/houdini/blob/main/app/javascript/components/tests/intl/index.tsx) with commented eslint modules.

jpcirqueira:
hello @wwahammy, my name is João from the University of Brasilia this is my first contribution on houdini. I know that the code reported by lucasgandac cannot be removed, but i will look for some other code that is not being used.

lucasA27:
I'm Lucas from the University of Brasilia and I will pair with @jpcirqueira on this issue and look for codes that are not used.

wwahammy:
Thanks @jpcirqueira and @lucasA27! Welcome to Houdini!

pedrohelias:
Hi @wwahammy, is this issue still open? Me and @GabriellyAssuncao are wanting to get this. Since the date of the last commentary, about 1 year, we think that can have useless code

PR #1000, opened 2022-02-09
Cleaning unused code

lucasA27:
## Description:

On some files in houdini, there are some unused code that might be deleted to improve code cleanliness.

Closes: https://github.com/houdiniproject/houdini/issues/798
Co-authored-by: João Pedro <jpcirqueira81@gmail.com>

Knowing the importance of clean code, we found it necessary to clean up unused code.

## Solution summary:
- Removed unused imports
- Removed unused variables
- Removed unnecessary async


app/javascript/components/form_fields/TextField.spec.tsx
```diff
@@ -43,7 +43,7 @@ describe('TextField', () => {
 
 		it('is valid when value is changed and then deleted', async () => {
 			expect.assertions(1);
-			await userEvent.type(emptyTextField, "enter some information{selectall}{backspace}");
```
wwahammy, line 46: Let's keep the await in. We have so many issues with inconsistent results on these tests when we don't use await.

Alternatively: figure out how to get the results to be more consistent :smile:

app/javascript/components/form_fields/TextField.spec.tsx
```diff
@@ -87,7 +87,7 @@ describe('TextField', () => {
 
 		it('is invalid when value is changed', async () => {
 			expect.assertions(1);
-			await userEvent.type(prefilledTextField, "{selectall}{backspace}");
```
wwahammy, line 90: See above

app/javascript/components/form_fields/TextField.spec.tsx
```diff
@@ -124,7 +124,7 @@ describe('TextField', () => {
 		it('is valid when value is changed to be longer than 10', async () => {
 			expect.assertions(1);
 			// We have to await here becuase validation takes a bit.
-			await userEvent.type(emptyTextField, 'this is longer than 10');
```
wwahammy, line 127: See above

app/javascript/legacy/settings/index/branding/index.js
```diff
@@ -1,9 +1,7 @@
 // License: LGPL-3.0-or-later
 // npm
-const snabbdom = require('snabbdom')
```
wwahammy, line 3: On all of these unused requires and imports, is there a way we can check for them as part of `eslint` or similar tool? That way we fix the problem for good.

wwahammy (changes requested):
Really good work all! I requested a few changes that make the work you did more generalizable. It's a little more work to get set up but we can guarantee we'll have less unused code over time.

app/javascript/components/form_fields/TextField.spec.tsx
jpcirqueira: sorry, we're back to 'await'. we removed await because we saw in the react testing library documentation that userEvent is a synchronous call, we could not reproduce the inconsistency.

app/javascript/components/form_fields/TextField.spec.tsx
jpcirqueira: desculpe, devolvemos o 'aguardo'.

app/javascript/legacy/settings/index/branding/index.js
jpcirqueira: The rule for this is 'no-unused-vars' is already being used, however several files are in eslintignore.

app/javascript/components/form_fields/TextField.spec.tsx
wwahammy: The error is pretty intermittent during the test. It's very frustrating.

app/javascript/legacy/settings/index/branding/index.js
wwahammy: Let's do this a different way then: could you change the eslint configuration so it skips most of the rules on these old ignored files but does run the `no-unused-vars`? That should have a minimal effect on the git history but will help.

lucasA27: Should a new issue be created for these changes? Because the purpose of the issue was to clean up code.

wwahammy: It's solving the same problem so I'm fine using the same issue and PR in this case.

wwahammy: Thanks for asking though!

jpcirqueira: we managed to find a solution for this by putting a new eslintrc file inside the legacy folder. Would this be an acceptable solution? if yes this can also be replicated in react_legacy

jpcirqueira: unfortunately we were not able to make an exception from the main rule only for legacy files, because apparently there is no way to deny file in the regex ref: https://github.com/selaux/eslint-plugin-filenames/issues/39. The solution described above solves the problem but it does not seem to be scalable, because we will always have to add new folders to the array, so this solution was not committed.

this is the root folder file :

image

this is the legacy folder file:

image

jpcirqueira: @wwahammy, do you know if this solution is acceptable? for us to make the commits

wwahammy: @jpcirqueira I think so? I'm not sure I fully understand it, could you make the change and push it so I can see what it looks like?

lucasA27: @wwahammy For this solution, @jpcirqueira and I added babel-eslint, to solve some parser errors. We also added overrides to legacy files to overwrite existing rules. There were 273 errors, do you think it makes sense to continue with this fix?

jpcirqueira:
@wwahammy  do you think we can continue with this solution? even with these 273 errors

app/javascript/legacy/settings/index/branding/index.js
wwahammy: @lucasA27 sorry about the delay in getting back to you, this slipped off my todo list. What was the reason that babel-eslint was needed?

jpcirqueira: @wwahammy sorry for the delay, without babel-eslint some "parse errors" were happening in the legacy files.

wwahammy:
Thanks for all the help @lucasA27 and @jpcirqueira. I think we'll tackle this a different way so I'm going to close this PR. Your help is much appreciated though!

PR #1001, opened 2022-02-10
Remove ";" from the login screen

KalebeLopes:
## Description:

Removal of a strange ";" lost on login screen.

Co-authored-by: Paulo Batista <matpaulo@hotmail.com.br>

Because of the need to have a clean login page with no code errors.

## Solution summary:

* Missing semicolon removed.

### DIFF
--- a/app/javascript/components/users/SignInComponent.tsx
+++ b/app/javascript/components/users/SignInComponent.tsx
@@ -254,7 +254,7 @@ function SignInComponent(props: SignInComponentProps): JSX.Element {
 				showProgressAndSuccess={showProgressAndSuccess}
 				signedIn={signedIn}
 				form={form}
-			/>;
+			/>
 		</form>
 	);
 }


app/javascript/components/users/SignInComponent.tsx
```diff
@@ -394,4 +394,4 @@ SignInComponent.defaultProps = {
 	onSubmitting: noop,
 };
 
-export default SignInComponent;
\ No newline at end of file
+export default SignInComponent;
```
wwahammy, line 397: What changed on this line?

wwahammy (commented):
I only have one question on the changes but otherwise it's a good catch!

app/javascript/components/users/SignInComponent.tsx
KalebeLopes: Apparently a newline at the end of the file. Could this become a problem?

wwahammy: Probably not. In the future, try to limit any extraneous changes in PRs. It can make it hard to track changes over time. Since it's very small though, I'll merge it this time.

PR #1002, opened 2022-02-10
Feat/meaningful methods names 783

GuilhermeBraz:
### Description:
Changed name of some methods to more meaningfull ones

Closes: https://github.com/houdiniproject/houdini/issues/783
Co-authored-by: @WeillerFernandes

To give more meaningful names to some methods present in the code, which did not have much clarity about their functionality in their original name. 

### Solution summary:
1. [simple_date](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L31) -> format_date_to_mdY
1. [readable_date](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L45) -> format_date_with_month_name
1. [url_with_param](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L59) -> append_param_to_url
1. [date_and_time](https://github.com/houdiniproject/houdini/blob/40debec3d8dab0dff27ca3c9d1b2a87350ca922d/app/helpers/application_helper.rb#L49) -> format_date_to_datetime_timezone
2. [finished?]( https://github.com/FGA-GCES/houdini/blob/8cac9dc1d31ce9804c561a32ce85b73a06b1b75e/app/models/campaign.rb#L175) -> deadline_has_been_reached
2. [montlhy_total](https://github.com/FGA-GCES/houdini/blob/8cac9dc1d31ce9804c561a32ce85b73a06b1b75e/app/models/recurring_donation.rb#L63)  -> calculate_monthly_donation_total 
6.1 [monthly_total](https://github.com/FGA-GCES/houdini/blob/8cac9dc1d31ce9804c561a32ce85b73a06b1b75e/app/legacy_lib/query_recurring_donations.rb) -> calculate_monthly_donation_total


app/models/campaign.rb
```diff
@@ -172,7 +172,8 @@ def days_left
     (end_datetime.to_date - Date.today).to_i
   end
 
-  def finished?
+  #The campaign has reached its end date time
+  def deadline_has_been_reached?
```
wwahammy, line 176: I don't see an advantage to this change. It seems to be more wordy without much benefit. I think `finished?` makes more sense.

app/helpers/application_helper.rb
```diff
@@ -56,7 +59,7 @@ def us_states
   end
 
   # Append a parameter to a URL string
-  def url_with_param(param, val, url)
+  def append_param_to_url(param, val, url)
```
wwahammy, line 62: Is this method used anywhere? There are WAY better ways to do this so, if it's not being used, the method should just be removed.

app/models/recurring_donation.rb
```diff
@@ -54,13 +54,14 @@ def total_given
...
   # Or just push it into the front-end
-  def self.monthly_total
-    all.map(&:monthly_total).sum
+  #Calculate the monthly donation total
+  def self.calculate_monthly_donation_total
+    all.map(&:calculate_monthly_donation_total).sum
   end
 
-  def monthly_total
+  def calculate_monthly_donation_total
```
wwahammy, line 64: These methods don't seem to be used in Houdini anywhere (and they're not very efficient code anyways). Let's just removed them.

app/views/campaigns/show.html.erb
```diff
@@ -88,7 +88,7 @@
 
     <% if !@campaign.child_campaign? %>
       <aside class='u-marginBottom--15 pastelBox--grey'>
-        <a class='button u-width--full' target='_blank' if-branded='background-color, dark' href='<%= @campaign.finished? ? "/peer-to-peer?npo_id=#{@nonprofit.id}" : "/peer-to-peer?campaign_id=#{ @peer_to_peer_campaign_param}" %>'>
+        <a class='button u-width--full' target='_blank' if-branded='background-color, dark' href='<%= @campaign.deadline_has_been_reached? "/peer-to-peer?npo_id=#{@nonprofit.id}" : "/peer-to-peer?campaign_id=#{ @peer_to_peer_campaign_param}" %>'>
```
wwahammy, line 91: As part of this pull request, since we're messing with it, can you replace these examples of hardcoded paths with rails path helpers? You can find out more about them in the guide on routing at the official Rails website.

app/helpers/application_helper.rb
wwahammy: In fact, the method doesn't seem to be used anywhere. Let's just remove it.

wwahammy (changes requested):
I like the work everyone did here. I have a few tweaks I'd like done before merging. Thanks so much for the help so far.

GuilhermeBraz:
Hey @wwahammy, thanks for the feedback! We re going to make these new changes then. Just one question, for the rail paths changes, you think that there is no need to create an issue and we should just fix within this pull request right?

wwahammy (approved):

WeillerFernandes:
Hey @wwahammy, in the last commit (53987dc) we removed the methods you asked for. Now we are studying the rails documentation to also make the rails path changes and be able to complete this PR, ok?

wwahammy:
Just a quick tip @WeillerFernandes, you can get all of the routes by running `bin/rails routes`. That might give you a start.

clarissalimab:
@WeillerFernandes could you provide more information on why you can't see that menu? Is that because you can't create a campaign or you can't see it because it's not a child campaign?

If it's the latter reason, you could create a regular campaign, and then on the campaign page that you created, you'd see a button saying something like "Start your own campaign for the nonprofit". There you'd create a child campaign and you'd be able to see the element that you're referring to.

As for the code you wrote, you're close, but I think it won't work. Assuming you're naming the paths correctly, you could do:

```ruby
href='<%= @campaign.finished? peer_to_peer_path(npo_id: @nonprofit.id) : peer_to_peer_path(campaign_id: @peer_to_peer_campaign_param) %>'>
```

clarissalimab:
Also, using the `link_to` function would be better than using the `<a>` element. This element would look like:

```ruby
<%= link_to '', @campaign.finished? peer_to_peer_path(npo_id: @nonprofit.id) : peer_to_peer_path(campaign_id: @peer_to_peer_campaign_param), class: 'button u-width--full', target: '_blank', 'if-branded' => 'background-color, dark' %>
```

I'm not sure about the syntax for the `if-branded` attribute, you could test to see if it works that way.

WeillerFernandes:
Thanks for the suggestions @clarissalimab. Regarding the problem with the campaign menu, when trying to create a new campaign, the program asks me to confirm my email first, but I didn't receive any email and I can't even ask for the confirmation link to be resent. Also, when trying to access the campaign menu I come across a routing error.

wwahammy:
@WeillerFernandes I think this means your user needs to be confirmed. Did you create run `houdini:nonprofit:create` with `--confirm-admin`?

WeillerFernandes:
@wwahammy I've already tried creating users in every way: with `--confirm-admin`, with `--no-confirm-admin` and without any arguments. In all cases I had the same problem.

wwahammy:
@WeillerFernandes I'm not quite sure what might be causing this bizarre behavior. Let's undo the path helper change and then I can review and merge the rest.

WeillerFernandes:
Okay @wwahammy, I undid the path helper change right now

wwahammy (approved):
Looks good!

wwahammy:
Wow, I approved it twice in a row :laughing:

Issue #661, opened 2021-08-17
[FEATURE] Change to using strict Typescript


wwahammy:
**Is your feature request related to a problem? Please describe.**
Typescript has valuable typechecks through it's [strict mode](https://www.typescriptlang.org/tsconfig#strict) that we're not using because we don't have strict mode on. Using `strict` for typescript may also allow usage of the strict Javascript mode which has increased performance in browser.

**Describe the solution you'd like**
* Add `strict: true` to `tsconfig.json`
* Fix all of the strict errors. Most of them should be straightforward.

GuilhermeBraz:
Hey @wwahammy, me and @JoaoVitorFarias are from University of Brasilia. We're excited to start contributing to the houdini project and be of help to the houdini community. It will be our first contribution!

PR #1012, opened 2022-02-19
Change to strict Typescript

GuilhermeBraz:
## Description:
- Added [strict: true](https://github.com/FGA-GCES/houdini/blob/b4e0d93c5e75352f764de846891bf6342e577e8f/tsconfig.json#L19)
-  Corrected the errors resulted of the change
- The complexity of the issue was harder than expected, the change to strict caused approximately 89 errors.
Closes: https://github.com/houdiniproject/houdini/issues/661
Co-authored-by: @JoaoVitorFarias

Using strict for typescript may also allow usage of the strict Javascript mode which has increased performance in browser.

## Solution summary:
- Added [strict: true](https://github.com/FGA-GCES/houdini/blob/b4e0d93c5e75352f764de846891bf6342e577e8f/tsconfig.json#L19)
Errors that were generated in the first test ( with links for their stackoverflow solutions):
- [TS2322](https://stackoverflow.com/questions/54496398/typescript-type-string-undefined-is-not-assignable-to-type-string)- 29 ocurrences
- [TS2564](https://stackoverflow.com/questions/49699067/property-has-no-initializer-and-is-not-definitely-assigned-in-the-construc)- 32 ocurrences 
- [TS2790](https://stackoverflow.com/questions/68476586/the-operand-of-a-delete-operator-must-be-optional-object-is-possibly-undefi)- 4 ocurrences
- [TS2345](https://stackoverflow.com/questions/70221980/understanding-ts2345-argument-of-type-typeof-mycomponent-is-not-assignable-t/70222087)- 8 ocurrences
- [TS2722](https://stackoverflow.com/questions/68371833/cannot-invoke-an-object-which-is-possibly-undefined-even-after-ensuring-it)- 2 ocurrences
- [TS2531](https://stackoverflow.com/questions/49431880/ts2531-object-is-possibly-null)- 3 ocurrences
- [TS2339](https://stackoverflow.com/questions/40068806/ts2339-property-does-not-exist-on-type) OR  [TS2339](https://stackoverflow.com/questions/57336678/strictnullchecks-causes-error-hasownproperty-does-not-exist-for-generic-types)- 2 ocurrences
- [TS2352](https://stackoverflow.com/questions/67223224/typescript-ts2352-type-could-be-instantiated-with-an-arbitrary-type-which-co) OR [TS2352](https://stackoverflow.com/questions/60218424/how-can-i-cast-or-change-the-generic-type-in-typescript) - 1 ocurrence
- [TS2769](https://stackoverflow.com/questions/58502778/how-to-repair-a-ts2769-no-overload-matches-this-call)- 1 ocurrence
- [TS2554](https://stackoverflow.com/questions/68270475/typescript-error-ts2554-expected-0-arguments-but-got-1-while-using-usereducer) OR [TS2554](https://stackoverflow.com/questions/62146182/typescript-is-giving-me-an-error-expected-0-arguments-but-got-1)- 5 ocurrences
- [TS2774](https://stackoverflow.com/questions/70004258/ts2774-this-condition-will-always-return-true-since-this-function-is-always-def) - 1 ocurrence
- [TS2344](https://stackoverflow.com/questions/63617344/how-to-satisfy-the-constraint-of-recordstring-unknown-with-interface) OR [TS2344](https://stackoverflow.com/questions/61735762/error-ts2344-type-tk-does-not-satisfy-the-constraint) - 1 ocurrence
- Some other errors ocurred as well during the correction of the first erros


app/javascript/hooks/useIntl.ts
```diff
@@ -16,7 +16,7 @@ export type IntlShape = ParentIntlShape & {
 	formatMoney(amount: Money, opts?: FormatMoneyOptions): string;
 };
 
-export const IntlContext = React.createContext<IntlShape>(null as IntlShape);
+export const IntlContext = React.createContext<IntlShape>(null as unknown as IntlShape);
```
wwahammy, line 19: Casting to unknown defeats the typechecker again. Sometimes its necessary but you should have a good justification.

app/javascript/legacy_react/api/api/UsersApi.ts
```diff
@@ -40,7 +40,7 @@ export class UsersApi {
         }
     }
 
-    private extendObj<T1, T2 extends T1>(objA: T2, objB: T2): T1|T2 {
+    private extendObj<T1 extends object, T2 extends T1>(objA: T2, objB: T2): T1|T2 {
```
wwahammy, line 43: Good job!

app/javascript/legacy_react/api/configuration.ts
```diff
@@ -1,6 +1,6 @@
 export class Configuration {
-    apiKey: string;
-    username: string;
-    password: string;
-    accessToken: string | (() => string);
+    apiKey!: string;
```
wwahammy, line 2: Can you show that these fields will never be null? If not, you shouldn't be using a non-null assertion.

app/javascript/legacy_react/app/loading_indicator.ts
```diff
@@ -32,7 +32,7 @@ class PageProgressBar {
   }
 
   private getElem(): Element {
-    return document.getElementById("pageProgressBar")
+    return document.getElementById("pageProgressBar")!
```
wwahammy, line 35: Are you sure this will never be null? I don't think that's a reasonable conclusion.

app/javascript/legacy_react/src/components/common/DefaultCloseButton.tsx
```diff
@@ -64,10 +64,10 @@ const states: { [state: string]: any } = {
 @observer
 export class DefaultCloseButton extends React.Component<DefaultCloseButtonProps, {}> {
   @observable
-  hovering: boolean
+  hovering: boolean = false;
```
wwahammy, line 67: Look into how the `@observable` decorator works. I don't believe this is the correct way to fix this.

app/javascript/legacy_react/src/components/common/LabeledFieldComponent.tsx
```diff
@@ -10,7 +10,7 @@ export interface LabeledFieldComponentProps
     labelText: string
     inError:boolean
     error?:string
-    inStickyError?:boolean
+    inStickyError?:boolean | any
```
wwahammy, line 13: What the reasoning here for allowing `inStickyError` to be `any`? It seems like it should be a boolean.

app/javascript/legacy_react/src/components/common/Modal.tsx
```diff
@@ -38,7 +38,7 @@ class Modal extends React.Component<ModalProps, {}> {
             {this.props.showCloseButton ?
               <Column colSpan={1} breakSize={'xs'}>
                 <div style={{ textAlign: 'right' }}>
-                  <DefaultCloseButton onClick={() => this.props.onClose()} />
+                  <DefaultCloseButton onClick={() => {if(this.props.onClose)this.props.onClose()}} />
```
wwahammy, line 41: This should be formatted on multiple lines but I think this is generally correct.

app/javascript/legacy_react/src/components/common/ProgressableButton.tsx
```diff
@@ -19,13 +19,13 @@ class ProgressableButton extends React.Component<ProgressableButtonProps, {}> {
   render() {
     let ourData: {title: string, disabled: boolean, prefix: JSX.Element|null}= {
       title:this.props.buttonText,
-      disabled:this.props.disabled,
+      disabled:this.props.disabled!,
```
wwahammy, line 22: Are you sure `this.props.disabled` will always be non-null? If not, you can't make that assertion.

app/javascript/legacy_react/src/components/common/layout.tsx
```diff
@@ -10,11 +10,11 @@ function arrayify<T>(items: T[]|T){
 
 type ClassNameableChildren = React.ReactElement<ClassNameable>[]|React.ReactElement<ClassNameable>
 
-export const TwoColumnFields: React.StatelessComponent<{children:ClassNameableChildren}> = (props) => {
+export const TwoColumnFields: React.FunctionComponent<{children:ClassNameableChildren}> = (props) => {
```
wwahammy, line 13: You can just change this to a normal function. Look into function components on the React website.

app/javascript/legacy_react/src/components/common/layout.tsx
```diff
@@ -10,11 +10,11 @@ function arrayify<T>(items: T[]|T){
...
     const children = arrayify(props.children)
     return <Row>
         {
-            children.map((i:React.ReactElement<ClassNameable>) => {
+            children.map((i:React.ReactElement<ClassNameable> | any) => {
```
wwahammy, line 17: `any` defeats the type checker. Why switch to `any`?

wwahammy (changes requested):
Thanks a lot for working on this! You made some good strides but there's some work that needs to be done. I want to encourage you to think a little more big picture about this. The warnings were solved but it was done in a way that often overrode the benefits of strict typing. The goal isn't to solve the warnings, the goal is to use strict typing and the warnings help us do that. I commented in a number of places where I think a different approach would be better or I had questions. I only marked one or a few of each example for improvement; you should apply that feedback to all the other unmarked examples as well.

Also, you'll notice that the build failed here. In particular, `yarn eslint` failed and `bin/webpack` failed. Those are commands you can run on your computer. Those will eventually have to pass, along with `yarn jest`, before this can be merged in the future.

All and all though, it's a really good start!

GuilhermeBraz:
Hey @wwahammy thanks for the detailed feedback. We are going to study more the errors and the typescript language to make better changes. We are new to typescript so we didnt have much inside on what some corrections did to the code, but we still had the intuition that some changes maybe counterproductive but we were hoping for your feedback on them. Once again thanks for the detailed review and if you have some sugestions on certain types of error lets us know :)

Issue #967, opened 2022-01-10
[FEATURE] Rename `subtransaction_payments` to `payments`


wwahammy:
Currently, each `Subtransaction` has many `subtransaction_payments`. This name is redundant and only really exists because we have a legacy entity already called Payment. For simplicity, let's just rename `subtransaction_payments` to `payments`.

clarissalimab:
That would make it easier to understand, I support it.

MaiconMares:
Hi, I and @alan-ms are gonna be working in this issue, but we have some doubts. We want to know if the problem requires that `subtransaction_payments` entity and all its dependencies have to be renamed or it's needed just to rename its name on the places that `subtransaction_payments` are called. Thank you for your attention!

PR #1013, opened 2022-02-20
Rename subtransaction_payments to payment

MaiconMares:
### Description:
It solves an ambiguity in the name of `subtransaction_payments` entity.

Closes: #967 
Co-authored-by: @alan-ms <alan.sousa@protonmail.com>

This name is redundant and only really exists because we have a legacy entity already called Payment.

### Solution summary:
In the solution was implemented an alias that makes possible to refer to `subtransaction_payments` entity as `payments`.

### DIFF
--- a/app/models/concerns/model/subtransactable.rb
+++ b/app/models/concerns/model/subtransactable.rb
@@ -15,6 +15,6 @@ module Model::Subtransactable
 		has_one :supporter, through: :trx
 		has_one :nonprofit, through: :trx
 
-		has_many :payments, through: :subtransaction
+		has_many :subtransaction_payments, through: :subtransaction
 	end
 end
--- a/app/models/concerns/model/subtransactable.rb
+++ b/app/models/concerns/model/subtransactable.rb
@@ -15,6 +15,6 @@ module Model::Subtransactable
 		has_one :supporter, through: :trx
 		has_one :nonprofit, through: :trx
 
-		has_many :subtransaction_payments, through: :subtransaction
+		has_many :payments, through: :subtransaction
 	end
 end
--- a/app/views/api/payments/index.json.jbuilder
+++ b/app/views/api/payments/index.json.jbuilder
@@ -2,8 +2,8 @@
 
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
-json.data @payments, partial: '/api/payments/subtransaction_payment',
-																					as: :subtransaction_payment
+json.data @subtransaction_payments, partial: '/api/subtransaction_payments/subtransaction_payment',
+																																				as: :subtransaction_payment
 
 json.current_page @payments.current_page
 json.first_page @payments.first_page?
--- a/app/views/api/payments/index.json.jbuilder
+++ b/app/views/api/payments/index.json.jbuilder
@@ -2,11 +2,11 @@
 
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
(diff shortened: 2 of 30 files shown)


wwahammy:
@MaiconMares sure, no problem. I would have before but I didn't realize you had wanted this reviewed. In the future, you can request a review by going to reviewers in the top right of the page and putting in my username.

spec/models/campaign_gift_purchase_spec.rb
```diff
@@ -73,23 +73,23 @@
...
 
-  let(:cgp_builder_to_id) do 
+  let(:cgp_builder_to_id) do
     {
       'id' => match_houid('cgpur'),
       'object' => 'campaign_gift_purchase',
       'type' => 'trx_assignment'
     }
   end
-
```
wwahammy, line 90: Please undo all these formatting changes. They definitely should be done at some point but it makes it hard to review the PR.

app/views/api/payments/index.json.jbuilder
```diff
@@ -2,11 +2,11 @@
 
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
-json.data @subtransaction_payments, partial: '/api/subtransaction_payments/subtransaction_payment',
-																																				as: :subtransaction_payment
+json.data @payments, partial: '/api/payments/subtransaction_payment',
```
wwahammy, line 5: I don't think this should be changed, the directory should still be `/api/subtransaction_payments`

app/models/concerns/model/subtransactable.rb
```diff
@@ -15,6 +15,6 @@ module Model::Subtransactable
 		has_one :supporter, through: :trx
 		has_one :nonprofit, through: :trx
 
-		has_many :subtransaction_payments, through: :subtransaction
+		has_many :payments, through: :subtransaction
```
wwahammy, line 18: I'll be honest, I'm surprised this passes the tests. I had thought you needed to add the class name here like you did in `Subtransaction`.

wwahammy (changes requested):
Only a few minor tweaks but it's almost ready to merge. Thanks for all the help!

spec/models/campaign_gift_purchase_spec.rb
MaiconMares: Solved!

app/views/api/payments/index.json.jbuilder
MaiconMares: Solved!

app/models/concerns/model/subtransactable.rb
MaiconMares: Solved!

app/views/api/payments/index.json.jbuilder
```diff
@@ -5,8 +5,8 @@
 json.data @subtransaction_payments, partial: '/api/subtransaction_payments/subtransaction_payment',
```
wwahammy, line 5: @payments is the name of the class variable being set in the `index` method of `app/controllers/api/payments_controller.rb` that contains all of the `SubtransactionPayment`s to be displayed so that's the one that should be used.

Also, why is there a trailing comma here? That looks worrisome.

wwahammy (changes requested):
Thanks for all the help! I had one correction and question requested.

wwahammy:
... how is this passing?

app/views/api/payments/index.json.jbuilder
MaiconMares: I fixed the variable name. Sorry for trailing comma, the real problem was an automatic rubocop correction that had added a lot of tabs, so there is a last parameter after the comma trailing that was impossible to see. Please review our contribution again and thank you for your time!!!

wwahammy (approved):
Looks good to me! Thanks for the help!

Issue #964, opened 2022-01-07
[FEATURE] Create `/api/nonprofit/:nonprofit_id/roles` for seeing the list of users with a role on a nonprofit


wwahammy:
Depends on #963 

Currently there's no way to see what roles and users that are associated with a nonprofit.

`/api/nonprofit/:nonprofit_id/roles` would list those roles.

wwahammy:
Thoughts @clarissalimab?

KalebeLopes:
Hi @wwahammy! @JoaoVitorFarias and i are students at the University of Brasilia and we are excited to contribute to this issue!

PR #1016, opened 2022-02-23
feat: create router for roles nonprofit

KalebeLopes:
### Description:
Currently there's no way to see what roles and users that are associated with a nonprofit.

Closes: #964
Co-authored-by: alan-ms <alan.sousa@protonmail.com>
Co-authored-by: JoaoVitorFarias <jvlopesfarias@gmail.com>

Currently, there's no way to know what roles a particular nonprofit has and it causes impacts on daily operations.

### Solution summary:
The endpoint "/api/nonprofit/:nonprofit_id/roles" was created, which returns the roles of a nonprofit through its id, now it is possible to know which roles a nonprofit has.

Screenshot from 2022-02-22 21-54-33

### DIFF
--- a/app/controllers/api/roles_controller.rb
+++ b/app/controllers/api/roles_controller.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
+class Api::RolesController < Api::ApiController
+	include Controllers::Nonprofit::Current
+	include Controllers::Nonprofit::Authorization
+
+	# before_action :authenticate_nonprofit_user!, except: %i[new create]
+
+	# get /nonprofit/:nonprofit_id/roles
+	def index
+		@roles = current_nonprofit.roles
+	end
+end
--- a/app/views/api/roles/index.json.jbuilder
+++ b/app/views/api/roles/index.json.jbuilder
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
+json.data @roles, partial: '/api/roles/role', as: :role
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -15,6 +15,7 @@ Rails.application.routes.draw do
   defaults format: :json do # they're APIs, you have to use JSON
     namespace :api do
       resources :nonprofits, only: [:create, :show] do
+        resources(:roles, only: [:index])
         resources :campaigns, only: [:show] do
           resources :campaign_gift_options, only: [:index, :show]
         end
@@ -80,7 +81,6 @@ Rails.application.routes.draw do
       put(:followup, on: :member)
       post(:create_offsite, on: :collection)
     end
-
     resources(:charges, only: [:index]) do
       resources(:refunds, only: %i[create index])
     end


wwahammy (commented):
I had a few minor comments. I would also like tests added if we decide to move forward with this specific design.

I am wondering though on whether this is the right design for making nonprofit roles available. In general, a principle going forward for Houdini is that data on an instance is effectively partitioned between nonprofits. One of the advantages of that is you could put a server handling API calls in front of multiple coordinated Houdini instances and route the api calls to a particular instance based solely on the nonprofit's id.

One nuance of that is that users don't fit into that paradigm, they have to be handled differently. In effect, you could treat users as an entirely separate system as part of this model. But where do roles go? I'm not sure they should go under nonprofits because that sort of breaks the mental model of all the data accessible and modifiable under `/api/nonprofits/:nonprofit_id` being solely related to that nonprofit. But then where else should they go? And I'm not quite sure on that. So I guess I'm open to suggestions on possible places to put roles into this model, including where you put them.

@clarissalimab, do you have any thoughts?

wwahammy:
@KalebeLopes, @clarissalimab and I talked and we'd like you to go ahead with this design. Please add routing and requests specs to verify that the code works correctly.

KalebeLopes:
Dear @wwahammy, we have completed the changes. Could you kindly take a look?

app/controllers/api/roles_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
+class Api::RolesController < Api::ApiController
+	include Controllers::Nonprofit::Current
+	include Controllers::Nonprofit::Authorization
+	# before_action :authenticate_nonprofit_user!, except: %i[new create]
```
wwahammy, line 8: There's no authentication happening here.

spec/controllers/api/roles_controller_spec.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
+require 'rails_helper'
+require 'controllers/support/shared_user_context'
+
+describe Api::RolesController, type: :controller do
```
wwahammy, line 8: Please use a request spec instead and verify authorization to access this data.

app/controllers/api/roles_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+	# get /nonprofits/:nonprofit_id/roles
+	def index
+		@roles = current_nonprofit.roles
```
wwahammy, line 12: Please use kaminari paging with this so it's consistent with other endpoints.

app/views/api/roles/index.json.jbuilder
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+# License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
+# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
+json.data @roles, partial: '/api/roles/role', as: :role
```
wwahammy, line 5: Please add the additional kaminari attributes like used in other endpoints.

wwahammy (changes requested):
Thanks for the work. There's a few additional things I'd like adjusted.

wwahammy:
This will be implemented in #1155

PR #1027, opened 2022-02-28
Correct specs broken by fastener and not caught by turbo_test

wwahammy:
Two bugs caused us to think tests were passing when they weren't. The problem was that we were using turbo_test which doesn't always return a non-zero error code. The intent
of using turbo_test is that we could go faster on a multicore machine but the github runners aren't multicore so it's pointless.

Additionally, the use of fastener in #975 broke some code that we didn't catch. This reverts most of the changes from that PR.

- Correct some bugs which weren't caught due to using turbo_test
- Use the correct 'rake spec' instead of turbo_test

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -66,7 +66,7 @@ jobs:
           bundle exec rake spec
       - run: bin/setup ci
       - name: run spec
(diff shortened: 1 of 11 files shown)


wwahammy:
@clarissalimab I will merge this when it passes as `main` is currently broken.

PR #1028, opened 2022-03-01
Add spring, spring-commands-rspec, spring-commands-rubocop

wwahammy:
Spring will help preload your rails files faster during development and testing by keeping some of the start up files loaded in-memory.﻿


wwahammy:
I'm finding this a little buggy so let's not merge.

PR #1036, opened 2022-03-09
 Change contrast ratios login page accessibility

higton:
Issue #8 

Some colors were not following the recommended contrast ratios from the WCAG guidelines.
So this pr adjust the text colors from the login page to improve legibility and accessibility for all sighted users.

[Link](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast) for more information about it.

### DIFF
--- a/app/javascript/components/users/SignInComponent.tsx
+++ b/app/javascript/components/users/SignInComponent.tsx
@@ -189,6 +189,14 @@ function SignInComponent(props: SignInComponentProps): JSX.Element {
 
 	//Styling - Material-UI
 	const useStyles = makeStyles((theme: Theme) => createStyles({
+		root: {
+			'& .MuiFormHelperText-root.Mui-error': {
+				color: "#dd1a0c",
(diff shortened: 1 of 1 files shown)


wwahammy:
Thanks for this contribution @higton. This is actually addressed in a later version of Material UI, or, as it's now called MUI. I would prefer to upgrade instead of fixing this here. If you REALLY, wanted to try that, you're welcome but I would expect it to be quite difficult.

wwahammy:
@higton, were you interested in trying that upgrade? I did a first pass on it a few months ago but I remember I got stuck at some point so I took a break and haven't gotten back to it. If you're interested, I can figure out where that code is to give you a starting point.

higton:
@wwahammy, Thank you for suggesting this for me, seems interesting. Unfortunately, right now, my focus is to try to fix some problems from another page first. Maybe when I have some more time available.

PR #1037, opened 2022-03-10
Rename -su option to -s

BenSturmfels:
Running `bin/rails houdini:nonprofit:create -su` as per README.md and the built in command help gives `ERROR: "rails create" was called with arguments ["-su"]`. It appears that short arguments need to be a single character, so updating to `-s` as discussed in #1032.


wwahammy:
Looks good @BenSturmfels! Thanks for your contribution!

PR #1039, opened 2022-03-10
Use debug gem as recommended by Rails and Ruby. Remove byebug and pry.

wwahammy:
[debug](https://github.com/ruby/debug) is the newest tool for Ruby debugging. It's officially supported by the Ruby language team, it's highly performant, recommended by the Rails team and supports integration with IDEs like VSCode.

This PR replaces our old debuggers, byebug and pry, with debug itself.


clarissalimab (approved):
bye byebug

Issue #524, opened 2021-03-08
Unable to insert middleware error when running v2 in prod.


tony-gutierrez:
## Description
Added env vars. 
ran `bin/rails server -e production`

## Expected Behavior
Server starts

## Actual Behavior
Got: 
```
	11: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/initializable.rb:32:in `run'
	10: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/initializable.rb:32:in `instance_exec'
	 9: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/application/finisher.rb:108:in `block in <module:Finisher>'
	 8: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/engine.rb:521:in `app'
	 7: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/engine.rb:521:in `synchronize'
	 6: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/engine.rb:524:in `block in app'
	 5: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/configuration.rb:94:in `merge_into'
	 4: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/configuration.rb:94:in `each'
	 3: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/configuration.rb:95:in `block in merge_into'
	 2: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/railties-6.1.3/lib/rails/configuration.rb:53:in `block in insert_before'
	 1: from /usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/actionpack-6.1.3/lib/action_dispatch/middleware/stack.rb:101:in `insert'
/usr/local/rvm/gems/ruby-2.7.2@fundraising/gems/actionpack-6.1.3/lib/action_dispatch/middleware/stack.rb:162:in `assert_index': No such middleware to insert before: ActionDispatch::Static (RuntimeError)
```

## Your Environment
* Ruby version: 2.7.2
[...]

Issue #1046, opened 2022-03-16
[BUG] Invalid route name, already in use: 'rails_service_blob'


BenSturmfels:
As @iRunner mentioned in [#524](https://github.com/houdiniproject/houdini/issues/524#issuecomment-808832810), the `bin/rails server --environment=production` fails for me with an error`Invalid route name, already in use: 'rails_service_blob'`. This running the recent `main` branch on a Debian Bullseye VM with Ruby 2.7.4 and Bundler 2.2.5.

Removing `config/routes.rb` lines 292-305 appears to fix the issue, allowing the server to start up. These lines appear to be nearly identical to `gems/activestorage-6.1.5/config/routes.rb` lines 3-17 except for a minor difference in the last line.

I'm also seeing the same on my local Guix operating system, so it doesn't appear to be specific to the above VM.

```
$ bin/rails server --environment=production
=> Booting Puma
=> Rails 6.1.5 application starting in production 
=> Run `bin/rails server --help` for more startup options
Exiting
Traceback (most recent call last):
	70: from bin/rails:9:in `<main>'
	69: from /home/debian/houdini/mygems/gems/bootsnap-1.11.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
	68: from /home/debian/houdini/mygems/gems/bootsnap-1.11.1/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
	67: from /home/debian/houdini/mygems/gems/railties-6.1.5/lib/rails/commands.rb:18:in `<main>'
	66: from /home/debian/houdini/mygems/gems/railties-6.1.5/lib/rails/command.rb:48:in `invoke'
	65: from /home/debian/houdini/mygems/gems/railties-6.1.5/lib/rails/command/base.rb:69:in `perform'
	64: from /home/debian/houdini/mygems/gems/thor-1.2.1/lib/thor.rb:392:in `dispatch'
	63: from /home/debian/houdini/mygems/gems/thor-1.2.1/lib/thor/invocation.rb:127:in `invoke_command'
	62: from /home/debian/houdini/mygems/gems/thor-1.2.1/lib/thor/command.rb:27:in `run'
	61: from /home/debian/houdini/mygems/gems/railties-6.1.5/lib/rails/commands/server/server_command.rb:135:in `perform'
[...]

PR #1054, opened 2022-03-17
Remove font_assets dependency

BenSturmfels:
From discussion in #524. May potentially break fonts, but is more important that the application can start up in production mode.


wwahammy:
Did you see any places where the fonts were broken?

BenSturmfels:
Unfortunately prod isn't working for me at all due to #1046 so I can't confirm whether fonts are affected.

BenSturmfels:
@wwahammy now that I have `--environment=production` up and running Ican now confirm that the OpenSans and FontAwesome fonts are loading successfully for me. Icons are showing up and fonts look good. Confirmed in the browser devtools that the font I'm seeing on-screen is indeed OpenSans. Screenshot attached. Does this give you enough confidence to merge this PR?

PR #1055, opened 2022-03-17
Add some draft production deployment docs.

BenSturmfels:
I couldn't see an documentation about how to deploy Houdini in production, so I've added some. I'm a Rails newbie, so I'm sure you'll correct me where I'm heading in the wrong direction. :)

### DIFF
--- a/docs/production_deployment.md
+++ b/docs/production_deployment.md
@@ -37,7 +37,7 @@ the web server process, but in production mode this should be done beforehand
 (`config.assets.compile = false`). To compile the assets, run:
 
 ```sh
-bin/rake assets:precompile
+bin/rails assets:precompile
 ```
 
 As mentioned above, this step is memory intensive and will likely fail on
--- a/docs/production_deployment.md
+++ b/docs/production_deployment.md
@@ -49,6 +49,11 @@ These assets are also not served in production mode by default
 served directly from Nginx as configured below. You can override this behavior
 by setting the environment variable `RAILS_SERVE_STATIC_FILES=true`.
 
+You may also consider serving [static assets from a
+CDN](https://guides.rubyonrails.org/asset_pipeline.html#cdns). This is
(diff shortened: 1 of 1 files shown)


wwahammy:
Is this ready for review and possible merging @BenSturmfels?

BenSturmfels:
@wwahammy yes, this should be ready to review.

docs/production_deployment.md
```diff
@@ -0,0 +1,118 @@
...
+
+```sh
+RAILS_ENV=production bin/rails assets:clobber
+RAILS_ENV=production bin/rails assets:precompile
+```
+
+As mentioned above, this step is memory intensive and will likely fail on
+systems with less than 3GB RAM free.
+
+These assets are also not served in production mode by default
```
wwahammy, line 47: It would be good to also mention that you can use a CDN for the assets, which is a common recommendation: https://guides.rubyonrails.org/asset_pipeline.html#set-up-a-cdn-to-serve-static-assets

wwahammy (commented):
I like it @BenSturmfels. It would be good to at least mention that you can deploy assets via a CDN and link to https://guides.rubyonrails.org/asset_pipeline.html#set-up-a-cdn-to-serve-static-assets.

With that super minor tweak, I'm all for merging!

docs/production_deployment.md
BenSturmfels: @wwahammy I've now added a note about CDN setup.

(To be honest, it's not an approach I use myself on web applications. For small-medium traffic sites I find that serving files from Nginx is low-overhead and operationally simple. For high traffic sites, I find that it's useful to put the whole site behind a CDN to provide both caching and some basic security for the origin server. Having the whole site behind the CDN also make it a little easier to troubleshoot since all the traffic is taking the same path.)

wwahammy (approved):
Looks great @BenSturmfels. Thanks for the help!

PR #1063, opened 2022-03-24
Feature/add method handling expansion child objects

### LINKED ISSUE #962: [FEATURE] Add a method for handling expansion of child objects in jbuilder
Currently, there's no way to expand child objects in jbuilder files. Ultimately, we should be able to do this. As an example, look at the following (simplified) transaction object for:

```jsonc

// TRANSACTION OBJECT 1

{
  "object": "transaction",
  "id": "trx_35n1235o",
  // other attributes hidden for simplicity
  "supporter": "supp_35135no",
  "subtransaction": "offlinetrx_t35h23o5"
}
```

A user may instead want the subtransaction expanded so they don't have to make a second request:

```jsonc

// TRANSACTION OBJECT 2

{
  "object": "transaction",
  "id": "trx_35n1235o",
  // other attributes hidden for simplicity
  "supporter": "supp_35135no",
  "subtransaction": {
    "id": "offlinetrx_t35h23o5",
    "object": "offline_transaction",
    "amount": {
      "cents": 4000,
      "currency": "usd"
    },
    // other attributes hidden for simplicity
    "subtransaction_payments": [
      {
        "id": "offtrxchrg_ewiothat",
        "object": "offline_transaction_charge",
        "type": "payment"
      }
    ]
  }
}
```

As user may want the supporter and the subtransaction expanded both:

```jsonc

// TRANSACTION OBJECT 3

{
  "object": "transaction",
  "id": "trx_35n1235o",
  // other attributes hidden for simplicity
  "supporter": {
    "id": "supp_35135no",
    "object": "supporter",
    "name": "Penelope Schultz",
    // other attributes hidden for simplicity
  },
  "subtransaction": {
    "id": "offlinetrx_t35h23o5",
    "object": "offline_transaction",
    "amount": {
      "cents": 4000,
      "currency": "usd"
    },
    // other attributes hidden for simplicity
    "subtransaction_payments": [
      {
        "id": "offtrxchrg_ewiothat",
        "object": "offline_transaction_charge",
        "type": "payment"
      }
    ]
  }
}
```

Additionally, a user may want the subtransaction payments expanded, but now the supporter:

```jsonc

// TRANSACTION OBJECT 4

{
  "object": "transaction",
  "id": "trx_35n1235o",
  // other attributes hidden for simplicity
  "supporter": "supp_35135no",
  "subtransaction": {
    "id": "offlinetrx_t35h23o5",
    "object": "offline_transaction",
    "amount": {
      "cents": 4000,
      "currency": "usd"
    },
    // other attributes hidden for simplicity
    "subtransaction_payments": [
      {
        "id": "offtrxchrg_ewiothat",
        "object": "offline_transaction_charge",
        "type": "payment",
        // other attributes hidden for simplicity
        "gross_amount": {
          "cents": 4000,
          "currency": "usd"
        },

        "net_amount": {
          "cents": 3700,
          "currency": "usd"
        },

        "fee_total": {
          "cents": 300,
          "currency": "usd"
        }
      }
    ]
  }
}
```

## A proposed solution

We should be able to pass an object describing the expansions to a call to render a jbuilder partial. I'll explain the object description and how the call should be used.

### Expansion descriptions

I believe a simple mechanism for expansion descriptions would be to provide an array that has the dot paths to parts of the JSON to expand from the root element. Here are the expansions for each of examples:

```javascript
//TRANSACTION OBJECT 1
obj_1_expansion = [] // nothing to expand

//TRANSACTION OBJECT 2
obj_2_expansion = ["subtransaction"] // expand the subtransaction from the root object

//TRANSACTION OBJECT 3
obj_3_expansion = ["subtransaction", "supporter"] // expand the subtransaction and supporter from the root object

//TRANSACTION OBJECT 4
obj_4_expansion = ["subtransaction", "subtransaction.subtransaction_payments"] // expand the subtransaction from the root object and then subtransaction_payments from the subtransaction object

//OR

obj_4_expansion_ideal = ["subtransaction.subtransaction_payments"] // Since you have to expand the subtransaction in order to expand the subtransaction_payment.

```

### Passing to a jbuilder template

From the level of the controller, you can pass the expansions by setting the `@__expand` variable. (Named like this to avoid any sort of collisions with something named `@expand`). Let's show how this would look in `TransactionController` if we always want to expand the supporter.

```ruby
class Api::TransactionsController < Api::ApiController

  # OTHER CONTENTS REMOVED FOR SIMPLICITY
	def show
		@transaction = current_transaction
    @__expand = ['supporter'] ## let's assume we want always want to expand the supporter
	end
end

```

#### Sanitizing expansion requests

We usually want to allow an API user to provide the specific expansions they want. On a transaction object, users may want supporter expanded while others may want the subtransaction expanded. To that end, we can have the user send a parameter named `__expand` which contains an array of dot paths to expand.

We probably should not allow an unlimited set of expansions, at least for all users. A user could request a really large set of expansions which might make the requests go slowly and overload the server. For example, let's say from the transaction element, request a super long expansion like: `subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction.subtransaction.subtransaction_payments.transaction`. That's a valid expansion but it's pointlessly long. Therefore, we should likely prevent an average user from expanding more than one level deep (perhaps super_admins could have more?).

We would create a method for controllers called `sanitized_expansions` which would:

* turns into `params['__expand']` into an containing the dot path to be expanded. This will also remove duplicates, for example. If they request:
`["supporter", "subtransaction", "subtransaction.subtransaction_payments"]`, they'll get an object representing a tree that looks, more or less, like this:

```jsonc
{
  "supporter": null, // no children so it's null
  "subtransaction": { // `subtransaction` and `subtransaction.subtransaction_payments` gets combined to `subtransaction.subtransaction_payments`
                      // since `subtransaction` needs to be expanded for `subtransaction.subtransaction_payments`
    "subtransaction_payments": null //no children so it's null
  }
}

```

* here we also remove any dangerous expansions, which is likely only ones which are too long.
* return the result

It would then could be used as follows:

```ruby
class Api::TransactionsController < Api::ApiController

  # OTHER CONTENTS REMOVED FOR SIMPLICITY
	def show
		@transaction = current_transaction
    @__expand = sanitized_expansions
	end
end

```

### passing from a template to a partial

Once you're in a jbuilder template, you need to pass the expansions to a partial, you can use it like shown in the `app/views/api/transactions/show.json.jbuilder`

```ruby
json.partial! @transaction, as: :transaction, __expand: @__expand ## TODO: I have to verify this works
```

### passing between partials

Once in a partial, you can add use the `handle_expansion` helper method (maybe a better name exists?) like shown in in this example `app/views/api/transactions/_transactions.json.jbuilder`

```ruby

json.(:transaction, :id)

json.created transaction.created.to_i

# irrelevant properties removed for clarity

handle_expansion(:supporter, transaction.supporter, {json:json, __expand:__expand})

```

`handle_expansion` here accepts:

* the attribute name, which is `supporter`
* the object which may or may not be expanded
* a hash with:
  * as: the variable for the object for passing into the partial, defaults to the attribute_name
  * json: the jbuilder object
  * the `__expand` variable (we might be able to get this automatically)

If `__expand` is not set to expand supporter, this is the equivalent of:

```ruby
json.supporter transaction.supporter.id
```

if `__expand` is set to expand supporter, this is the equivalent of:

```ruby
json.supporter do
  json.partial! transaction.supporter, as: :supporter, __expand: __expand.children_of('supporter') # children_of gets the part of the tree below supporter, in this case, an empty tree
end
```

For array elements, we use the `handle_array_expansion` and `handle_item_expansion` methods. Let's assume we're in the `subtransaction` partial now and this is for `subtransaction_payments`. we would use it as follows:

```ruby

handle_array_expansion(:payments, subtransaction.subtransaction_payments, {json:json, __expand:__expand}) do |payment, opts|
  handle_item_expansion(payment, {json:opts.json, as: opts.item_as, __expand: opts.__expand})
end

```

`handle_array_expansion` here accepts:

* the attribute name, in this case `payments`
* the object decide on how to expand, in this case `subtransaction.subtransaction_payments`
* as hash with:
  * json: the jbuilder object
  * the `__expand` variable (we might be able to get this automatically)
  * the `item_as`: the name of the variable for passing the item into `handle_item_expansion`

* a block for displaying the item which has two parameters:
  * the array item 
  * an opts hash which contains:
    * `json`: the jbuilder object
    * `__expand`: which is the result of `__expand.children_of(attribute_name)` as passed into the object
    * `as`: the name of the variable when passed into a partial for json

`handle_item_expansion` accepts:

* the object which may or may not be expanded
* a hash with:
  * `json`: the jbuilder object
  * `as`: the name of the variable when passed into a partial
  * `__expand`: the expand variable for what can be expanded in the partial

#### Result

If payments is not supposed to be expanded, this is the equivalent of:

```ruby

json.payments subtransaction.subtransaction_payments do |payment|
  json.id payment.id
  json.object 'offline_transaction_charge'
  json.type 'payment'
end
```

If payments is supposed to be expanded this is the equivalent of:

```ruby
json.payments subtransaction.subtransaction_payments do |payment|
  json.id payment.id
  json.object 'offline_transaction_charge'
  json.type 'payment'
  json.supporter payment.supporter.id
  # removed other attributes for simplicity
end
```

If `payments.supporter` is supposed to be expanded, this is the equivalent of:

```ruby
json.payments subtransaction.subtransaction_payments do |payment|
  json.id payment.id
  json.object 'offline_transaction_charge'
  json.type 'payment'
  json.supporter do
    json.id payment.supporter.id
    json.object 'supporter'
    json.name payment.supporter.name
    # removed other attributes for simplicity
  end
  # removed other attributes for simplicity
end

```

alan-ms:
### Description
It addresses the problem when we want to expand child objects in jbuilder.

Closes #962 
Co-authored-by: MaiconMares [maiconmaresunb@gmail.com](mailto:maiconmaresunb@gmail.com)

###Motivation for implementing it:
At the moment we don't have a way to expand child objects in jbuilder, so we implemented an approach where this is possible.

###Solution summary:
Modified json partial d transaction, adding expansion of child objects.

```{
  "id": "1",
  "object": "transaction",
  "created": 1648087371,
  "supporter": {
    "id": 1,
    "name": null,
    "organization": null,
    "phone": null,
    "anonymous": false,
    "deleted": false,
    "object": "supporter",
    "merged_into": null,
    "supporter_addresses": [
      1
    ],
    "url": "http://localhost:3000/api/nonprofits/1/supporters/1",
    "nonprofit": 1
  },
  "nonprofit": 1,
  "amount": {
    "cents": 10,
    "currency": "usd"
  },
  "subtransaction": {
    "type": "subtransaction",
    "supporter": 1,
    "nonprofit": 1,
    "transaction": "1",
    "payments": [
      {
        "type": "payment",
        "id": "1",
        "supporter": 1,
        "nonprofit": 1,
        "transaction": "1",
        "subtransaction": {
          "id": "1",
          "object": "offline_transaction",
          "type": "subtransaction"
        },
        "object": "offline_transaction_charge",
        "net_amount": {
          "cents": 100,
          "currency": "usd"
        },
        "gross_amount": {
          "cents": 100,
          "currency": "usd"
        },
        "fee_total": {
          "cents": 100,
          "currency": "usd"
        },
        "created": 0,
        "url": "http://localhost:3000/api/nonprofits/1/transactions/1/subtransaction/payments/1"
      }
    ],
    "url": "http://localhost:3000/api/nonprofits/1/transactions/1/subtransaction",
    "object": "offline_transaction",
    "created": 1648087371,
    "amount": {
      "cents": 100,
      "currency": "usd"
    },
    "net_amount": {
      "cents": 100,
      "currency": "usd"
    }
  },
  "transaction_assignments": [
    {
      "type": "trx_assignment",
      "id": "1",
      "supporter": 1,
      "nonprofit": 1,
      "transaction": "1",
      "object": "donation",
      "designation": null,
      "amount": {
        "cents": 100,
        "currency": "usd"
      }
    }
  ],
  "url": "http://localhost:3000/api/nonprofits/1/transactions/1"
}```

### CODE REVIEW

**Comment by MaiconMares** [comment]
@wwahammy can you review our Pull Request and give feedbacks or guidances on what we need to improve/fix?

File: app/views/api/stripe_transaction_dispute_reversals/_stripe_transaction_dispute_reversal.json.jbuilder [hunk 1]
```diff
@@ -0,0 +1,18 @@
...
+	json.partial! '/api_new/common/amount', amount: paymentable.net_amount_as_money
+end
+
+json.gross_amount do
+	json.partial! '/api_new/common/amount', amount: paymentable.gross_amount_as_money
+end
+
+json.fee_total do
+	json.partial! '/api_new/common/amount', amount: paymentable.fee_total_as_money
+end
```
wwahammy, line 18: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transaction_dispute_reversals/object_events/_base.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transaction_disputes/_stripe_transaction_dispute.json.jbuilder
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transaction_disputes/object_events/_base.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transaction_refunds/_stripe_transaction_refund.json.jbuilder
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transaction_refunds/object_events/_base.json.jbuilder
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/stripe_transactions/_stripe_transaction.json.jbuilder
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
wwahammy, line 1: I don't believe this file has anything to do with the specific change.

app/views/api/subtransactions/_subtransaction.json.jbuilder
```diff
@@ -2,17 +2,17 @@
 
 # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
-json.(subtransaction.subtransactable, :id)
```
wwahammy, line 5: I think this should be kept for now.

wwahammy (changes requested):
Thanks for the work.

I think it'd be helpful to step back for a second and consider what the problem you're trying to solve is. How do all of the changes you've made help you do the following:

1. Add support for handling the expansion of child objects and
2. Prove that support works as expected.

A lot of these files here don't actually assist in that effort and therefore, they shouldn't be included. I believe there is testing around simple objects and that expansions work correctly in the source tree in CommitChange. I would recommend starting with that for your testing.

To be clear, you don't need to even use the new expansion code for all of the endpoints in this PR. In fact, it'd probably be better if you didn't!

app/views/api/stripe_transaction_dispute_reversals/_stripe_transaction_dispute_reversal.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transaction_dispute_reversals/object_events/_base.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transaction_disputes/_stripe_transaction_dispute.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transaction_disputes/object_events/_base.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transaction_refunds/_stripe_transaction_refund.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transaction_refunds/object_events/_base.json.jbuilder
MaiconMares: Done!

app/views/api/stripe_transactions/_stripe_transaction.json.jbuilder
MaiconMares: Done!

MaiconMares:
Hi @wwahammy I and @alan-ms have removed the unrelated files and refine the solution, we think the problem is solved.
**Obs.**: however the test from CommitChange doesn't work with our solution. We've thought it should work for our solution due to your recommendation, but it doesn't fit at all.

wwahammy:
Thanks @MaiconMares! When you say it "doesn't fit at all", what do you mean?

PR #1064, opened 2022-03-24
Switch to using javascript_packs_with_chunks_tag

### LINKED ISSUE #1060: [BUG] Stuck "Loading..." overlay and JS error "Stripe is not defined" when assets built with `NODE_ENV=production`
On the `main` branch, when I pre-build the assets with `rm -rf public/packs && bin/rake assets:precompile`, the Houdini dashboard get stuck with a greyed out "Loading..."  overlay and the following JS console errors:

```
ReferenceError: Stripe is not defined
ReferenceError: appl is not defined
Uncaught ReferenceError: I18n is not defined
```

If instead I `rm -rf public/packs` and run up the development server and let it build the assets, the dashboard loads with no issues. Similarly no issues if I `rm -rf public/packs && NODE_ENV=development bin/rake assets:precompile`. The difference between the 

The seems independent of whether I'm running the development server or the production server (with removed font-assets and `config.reload_routes = false` as per #1046 and #1054 to allow production to start).

The `page__strip_wrapper` JS file where `Stripe` is defined *is* being loaded, as shown in the above screenshot and it seems to be loaded before the error occurs. It seems that either the `page__strip_wrapper` JS isn't running or is being broken somehow by the production Webpack build process. I tried adding a `console.log('Hello from Stripe!')` at the bottom of `app/javascript/legacy/stripe_wrapper/index.es6` and rebuilding. The console message showed up for the inline development build, but not for the `assets:precompile` build.

I found that removing the `environment.SplitChunks` block from `config/webpack/production.js` and building with `rm -rf public/packs && bin/rake assets:precompile` also works, so the issue would appear to be related to chunking.

wwahammy:
We were incorrectly using javascript_pack_tag in places where javascript files were being split into chunks by webpacker. In that case, we should use javascript_pack_with_chunk_tags. This change does so.﻿

Closes #1060

### DIFF
--- a/app/views/campaigns/peer_to_peer.html.erb
+++ b/app/views/campaigns/peer_to_peer.html.erb
@@ -11,7 +11,7 @@
 
 <% content_for :javascripts do %>
 
-  <%= javascript_pack_tag 'i18n', 'page__campaigns__peer_to_peer' %>
+  <%= javascript_packs_with_chunks_tag 'i18n', 'page__campaigns__peer_to_peer' %>
 
   	<script>
     <% if @nonprofit %>
--- a/app/views/campaigns/show.html.erb
+++ b/app/views/campaigns/show.html.erb
@@ -24,7 +24,7 @@
 
   <%= render 'schema', campaign: @campaign, url: @url %>
   <%= render 'common/froala' if current_campaign_editor? %>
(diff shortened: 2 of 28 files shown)


BenSturmfels:
Thank you!

Issue #1050, opened 2022-03-16
[BUG] Stripe "You did not provide an API key" on donate form payment


BenSturmfels:
After configuring my local development environment with `STRIPE_API_KEY` and `STRIPE_API_PUBLIC` and adding some `puts` into `gems/bess/lib/houdini/engine.rb` to verify tha the environment variables are being being read in, I'm seeing the a "You did not provide an API key" error on the donate form (screenshot below).

![2022-03-17_10-01](https://user-images.githubusercontent.com/9541562/158705840-83cbf8fb-dc30-4fa7-922a-a55d9ecb128a.png)

From what I can see in the browser console, the `Authorization: Bearer` header is missing the the API key on the end, e.g. `Authorization: Bearer XXX`.

This is running against master in development mode, Ruby 2.7.4 and allowing assets to be built in-process (not a separate webpack process).

BenSturmfels:
(The Internal Server Error in my console are unrelated errors loading images caused by missing `libvips.so.42`)

BenSturmfels:
If I modify `bess` like this:

```patch
diff --git a/gems/bess/lib/houdini/engine.rb b/gems/bess/lib/houdini/engine.rb
index ddb499f4..fc6ef5b0 100644
--- a/gems/bess/lib/houdini/engine.rb
+++ b/gems/bess/lib/houdini/engine.rb
@@ -34,6 +34,8 @@ module Houdini
     config.houdini.payment_providers.stripe = ActiveSupport::OrderedOptions.new
     config.houdini.payment_providers.stripe.public_key = ENV['STRIPE_API_PUBLIC']
     config.houdini.payment_providers.stripe.private_key = ENV['STRIPE_API_KEY']
+    puts config.houdini.payment_providers.stripe.private_key
+    puts config.houdini.payment_providers.stripe.public_key
     config.houdini.payment_providers.stripe.connect = false
     config.houdini.payment_providers.stripe.proprietary_v2_js = false
```

and start `bin/rails server`, I see my Stripe keys printed to the console.

Then if I modify `_javascripts.html.erb` like this:

```patch
diff --git a/app/views/layouts/_javascripts.html.erb b/app/views/layouts/_javascripts.html.erb
index 5c724dc5..4c8ba88d 100644
--- a/app/views/layouts/_javascripts.html.erb
+++ b/app/views/layouts/_javascripts.html.erb
@@ -8,6 +8,9 @@
[...]

PR #1065, opened 2022-03-24
Fix Payment Provider Loading

wwahammy:
Depends on #1064

Fixes loading information for the `Houdini::PaymentProvider`'s

Closes #1050

PR #1066, opened 2022-03-25
Fix donations for Houdini

wwahammy:
There were a number of interconnected issues with the donation system. This is a quick fix for them but there are additional ones I am sure.

This seems to work on the pretty path but not quite sure how badly it works when there are errors

Depends on #1065
Depends on #1064


BenSturmfels:
Confirming that this fixes #1050! It doesn't fix #1067 or #1069.

clarissalimab (approved):
:rocket:

wwahammy:
@clarissalimab, I'm rerequesting review because I made a few more changes. I doubt it has any real effect on the end result but please look them over since I *THINK* this is passing... we'll see in a few minutes i guess!

wwahammy:
:fireworks:

clarissalimab (approved):
It looks good again!

BenSturmfels:
Brilliant! Thanks @clarissalimab and @wwahammy for your work on this!

PR #1070, opened 2022-03-25
Replace references to "rake" with "bin/rails"

BenSturmfels:
In #1060, @wwahammy mentioned in https://github.com/houdiniproject/houdini/issues/1060#issuecomment-1078418316 that this is recommended approach as of ~ Rails 6.

I've confirmed that:
* `bin/webpack` and `bin/webpack-dev-server` still run
* `script/compile-assets.sh` runs
* `script/db_total_reset.sh` runs
* JS tests still pass

I haven't tested:
* `Procfile.dev`
* `script/crontab.sh` (don't have rvm installed)


wwahammy (approved):
Looks good to me! Thanks for the PR!

PR #1082, opened 2022-04-01
Correct bugs in houdini:nonprofit:create and flip the meaning of confirm_admin

### LINKED ISSUE #1079: [BUG] "Create first campaign" link fails with AuthenticationError
On the `main` branch running in development, after creating an org and user with `bin/rails houdini:nonprofit:create --super-admin --confirm-admin --nonprofit-name=...`, when I click the "Create first campaign" list item on the dashboard home page, I get a 500 error:

```
method=GET path=/nonprofits/1/campaigns/name_and_id format=json controller=CampaignsController action=name_and_id status=500 error='AuthenticationError: AuthenticationError' duration=8.99 view=0.00 db=1.14 time=452686.65 exception=["AuthenticationError", "AuthenticationError"] exception_object=AuthenticationError
  
AuthenticationError (AuthenticationError):
  
app/controllers/concerns/controllers/user/authorization.rb:46:in `block (2 levels) in authenticate_confirmed_user!'
app/controllers/concerns/controllers/user/authorization.rb:45:in `authenticate_confirmed_user!'
app/controllers/concerns/controllers/locale.rb:22:in `switch_locale'
```

This seems to be because the user isn't confirmed. I've tried both `bin/rails houdini:nonprofit:create --super-admin --confirm-admin` and `bin/rails houdini:nonprofit:create --super-admin --no-confirm-admin` and neither of these seems to confirm the user. The CLI help says:

```
     [--confirm-admin], [--no-confirm-admin]  # Require the nonprofit admin to be confirmed via email
                                                                     # Default: true
```

That's actually a bit confusing to me. Should it *require* the user to be confirmed, or *confirm* the user immediately - it's not clear to me. I'd suggest removing the `--no-confirm-admin` option and changing the wording to be something like:

```
     [--confirm-admin]  # Skip the email confirmation step for this user
```

For now I'm manually confirming users with something like:

`update users set confirmed_at = '2022-03-30';`

After this, I can successfully create a campaign.

wwahammy:
Closes #1079

@BenStrumfels and @WeillerFernandes noticed that the documentation for houdini:nonprofit:create doesn't match the behavior.

The change does the following:

* Correctly sets super_admin status of the admin when `--super-admin` is used
* Changes the meaning of `--confirm-user` to autoconfirm the user on creation and sets it to true by default
* Correct the autoconfirmation of the user on creation

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ There are available arguments that add configurations on the nonprofit's creatio
 
 ```bash
   -s, [--super-admin], [--no-super-admin]     # Make the nonprofit admin a super user (they can access any nonprofit's dashboards)
-      [--confirm-admin], [--no-confirm-admin]  # Require the nonprofit admin to be confirmed via email
+      [--confirm-admin], [--no-confirm-admin]  # Autoconfirm the admin instead of waiting for them to click the email link
                                                # Default: true
 ```
 
--- a/app/legacy_lib/nonprofit_creation.rb
+++ b/app/legacy_lib/nonprofit_creation.rb
@@ -6,7 +6,7 @@
 #
 # NOTE: this should be moved to bess when Nonprofit and wiki is
 class NonprofitCreation
-	def initialize(result, options = nil)
+	def initialize(result, options = {})
 		result = sanitize_optional_fields(result)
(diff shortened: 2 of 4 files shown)


clarissalimab (approved):
Looks good to me!

Issue #1103, opened 2022-04-20
[BUG] Refund fails with HTTP 500


BenSturmfels:
Me again! ;) We have Houdini set up with a Stripe testing API key and I hit the "Refund" button for a payment, but it failed, showing `Error: cannot POST /nonprofits/1/charges/3/refunds (500)`.

Here's the output from the console. Appears to be an issue in `refunds_controller.rb`.
```

   (1.7ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
* Locale set to 'en'
  User Load (1.4ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/controllers/concerns/controllers/user/authorization.rb:71:in `administered_nonprofit'
   (1.3ms)  SELECT host_id FROM roles WHERE ("user_id" IN (1)) AND (roles.name IN ($Q$nonprofit_admin$Q$, $Q$nonprofit_associate$Q$))
  ↳ app/legacy_lib/query_roles.rb:19:in `host_ids'
  Nonprofit Load (1.1ms)  SELECT "nonprofits".* FROM "nonprofits" WHERE "nonprofits"."id" = $1 ORDER BY "nonprofits"."id" DESC LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/controllers/concerns/controllers/user/authorization.rb:73:in `administered_nonprofit'
  Nonprofit Load (0.5ms)  SELECT "nonprofits".* FROM "nonprofits" WHERE "nonprofits"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
  ↳ app/legacy_lib/fetch_nonprofit.rb:10:in `with_params'
   (0.4ms)  SELECT host_id FROM roles WHERE ("user_id" IN (1)) AND (roles.name IN ($Q$nonprofit_admin$Q$, $Q$nonprofit_associate$Q$))
  ↳ app/legacy_lib/query_roles.rb:19:in `host_ids'
  CACHE Nonprofit Load (0.0ms)  SELECT "nonprofits".* FROM "nonprofits" WHERE "nonprofits"."id" = $1 ORDER BY "nonprofits"."id" DESC LIMIT $2  [["id", 1], ["LIMIT", 1]]
[...]

PR #1107, opened 2022-04-21
Fix charge_params used during refund

BenSturmfels:
This change modifies `charge_params` to fix the `ArgumentError` mentioned in #1103 and to return a hash of the shape expected by `create`. I've used a symbol for `:refund` for consistency with `:charge_id` being a symbol. Would it be better to split these out into two functions though? Seems unclear to call the same function twice.

I've also removed the line setting `:user_id` because I can't see how it it does anything, though I may be unaware of some magic here.

I'm not very familiar with Ruby/Rails, but I'm keen to learn. :)

### DIFF
--- a/app/controllers/nonprofits/refunds_controller.rb
+++ b/app/controllers/nonprofits/refunds_controller.rb
@@ -11,23 +11,15 @@ module Nonprofits
 
     # post /charges/:charge_id/refunds
     def create
-      charge = Qx.select('*').from('charges').where(id: charge_params[:charge_id]).execute.first
-      render_json { InsertRefunds.with_stripe(charge, charge_params[:refund]) }
+      charge =  current_nonprofit.charges.find(params[:charge_id])
+      charge_params = params.require(:refund).permit(:amount).merge(user_id: current_user.id)
+      render_json { InsertRefunds.with_stripe(charge, charge_params) }
     end
 
     def index
       charge = current_nonprofit.charges.find(params[:charge_id])
       @refunds = charge.refunds
       render locals: {refunds: @refund}
-    end
-
-private
-
-    def charge_params
-      {
-        :charge_id => params.require(:charge_id),
-        :refund => params.require(:refund).permit(:amount)
-      }
     end
   end
 end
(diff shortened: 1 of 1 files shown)


app/controllers/nonprofits/refunds_controller.rb
```diff
@@ -12,8 +12,7 @@ class RefundsController < ApplicationController
     # post /charges/:charge_id/refunds
     def create
```
wwahammy, line 13: Goodness, this is not good code to start with. It was inherited sadly! Could you try this method with:

```suggestion
    def create
      charge =  current_nonprofit.charges.find(params[:charge_id])
      render_json { InsertRefunds.with_stripe(charge, charge_params.merge(user_id: current_user.id) }
    end
```

app/controllers/nonprofits/refunds_controller.rb
```diff
@@ -25,7 +24,10 @@ def index
 private
 
     def charge_params
-      params.require(:charge_id, refund: [:amount])
+      {
+        :charge_id => params.require(:charge_id),
+        :refund => params.require(:refund).permit(:amount)
+      }
```
wwahammy, line 30: You'll likely also need this change.

```suggestion      
  params.require(refund: :amount)
```

app/controllers/nonprofits/refunds_controller.rb
wwahammy: If line 15 doesn't work with the other suggested change on line 27, please try it with 

`charge_params[:refunds].merge(user_id: current_user.id)`

My mind is struggling to remember which was it's supposed to be :smile:

wwahammy (changes requested):
@BenSturmfels: First off, thanks for all the help! And significant parts of this project are not good examples of how to use Ruby and Rails sadly. This would be a good example of such a place.

app/controllers/nonprofits/refunds_controller.rb
BenSturmfels: I've now updated the `create` method and inlined the unnecessary extra `charge_params` method. The `require`/`permit` methods are making more sense to me now - I gather that these are used to clean the incoming HTTP POST data to make sure it has the items you expect and only those.

wwahammy:
@BenSturmfels just wanted to verify, have you tested that your code works? I assume it does but I wanted to verify :smile:

BenSturmfels:
> @BenSturmfels just wanted to verify, have you tested that your code works? I assume it does but I wanted to verify 😄

Thanks for checking - yes I've run this code and verified that it works for me.

PR #1113, opened 2022-04-26
Remove unused NonprofitsController#create

wwahammy:
We only allow Nonprofits to be created via CLI so we shouldn't be allowing people to register a nonprofit via an API.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -23,7 +23,7 @@ Rails.application.routes.draw do
           resources :ticket_levels, only: [:index, :show]
         end
         resources :supporters, only: [:index, :show] do 
-          resources :supporter_addresses, only: [:index, :show]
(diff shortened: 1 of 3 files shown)


config/routes.rb
```diff
@@ -23,7 +23,7 @@
           resources :ticket_levels, only: [:index, :show]
         end
         resources :supporters, only: [:index, :show] do 
-          resources :supporter_addresses, only: [:index, :show]
+          resources :supporter_addresses, only: [:index, :show]s
```
clarissalimab, line 26:
```suggestion
          resources :supporter_addresses, only: [:index, :show]
```

clarissalimab (approved):
Looks good! I just added a suggestion to remove a typo.

wwahammy:
@clarissalimab  Thanks for the correction!

PR #1115, opened 2022-04-27
Remove locale from email donor job

BenSturmfels:
payment_mailer.rb:22 does not pass the required locale to the `PaymentNotificationEmailDonorJob`, which creates an error when GoodJob attempts to run the task. To fix this I've removed the locale argument. An alternative would be to provide the locale, but I note that `resend_admin_receipt` similarly doesn't send a locale argument to `PaymentNotificationEmailNonprofitJob`, so the current approach is consistent.

This job error mentioned is currently silently swallowed in development - the email just doesn't go out.

### DIFF
--- a/app/jobs/payment_notification_email_donor_job.rb
+++ b/app/jobs/payment_notification_email_donor_job.rb
@@ -1,6 +1,6 @@
 class PaymentNotificationEmailDonorJob < EmailJob
 
-  def perform(donation, locale)
-    DonationMailer.donor_payment_notification(donation.id, locale).deliver_now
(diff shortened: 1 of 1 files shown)


wwahammy:
@BenSturmfels Thanks for this PR, unfortunately, we need the PaymentNotificationEmailDonorJob to get the locale; we expect there may be different translations of emails being sent. I'll look into why payment_mailer isn't passing the locale in. Did you report that as a bug?

BenSturmfels:
Thanks @wwahammy, I've now opened issue #1139 relating to this.

PR #1118, opened 2022-04-28
Document the process of confirming payments

BenSturmfels:
See discussion in #1102.

### DIFF
--- a/docs/payments.md
+++ b/docs/payments.md
@@ -1,0 +1,27 @@
+# Payments
+
+After configuring `STRIPE_API_KEY` and `STRIPE_API_PUBLIC` as per [Getting
+Started](getting_started.md), you should be able to successfully process credit
+card payments.
+
+## Confirming payments
+
+After a payment has been processed, Stripe consider the transaction to be
+"pending". Under normal circumstances, Stripe will update the status to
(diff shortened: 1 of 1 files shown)


docs/payments.md
```diff
@@ -0,0 +1,27 @@
...
+
+*Note*: Currently the `update_np_balances` job will fail silently unless you
+additionally configuring the `stripe_account_id` field for your `nonprofit`
+database record. For example:
+
+```sh
+$ psql houdini_production
+# UPDATE nonprofits SET stripe_account_id = 'acct_xxxxxxxxxxxxxxxx'
+  WHERE name = 'My Org';
+```
```
wwahammy, line 27: Best to avoid using SQL unless required. I would recommend the following:

```suggestion
$ bin/rails console
# Nonprofit.find_by(name: 'My Org').update(stripe_account_id: 'acct_xxxxxxxxxxxxxxxx')
```

wwahammy (changes requested):
Thanks @BenSturmfels. Normally, it's best to avoid using SQL unless there's no other option. Please consider the suggestion below.

wwahammy (approved):
Looks good!

PR #1119, opened 2022-04-28
Fix conflict with `acct` variable renaming

BenSturmfels:
The `acct` variable wasn't resolved properly when conflicts in this method were merged. One called it `acct` the other renamed to `account`.


clarissalimab (approved):
Nice catch!

PR #1171, opened 2022-07-05
Deprecate the use of querystring in app/javascript/legacy/nonprofits/supporters/index/sidepanel/index.js

wwahammy:
querystring isn't needed anymore as `URLSearchParams` exists. I've created an and added an implementation using USP and created specs for verifying it's actually compatible.


wwahammy:
@clarissalimab I think this is relatively simple change but I think it'd make sense to have you look it over.

wwahammy:
Since this has been out for review awhile, I'm going to merge it!

PR #1243, opened 2022-07-18
Removed offensive term from project

### LINKED ISSUE #354: [BUG] NPM dependency `validator` uses an offensive term as part of its API
## Description
`validator`, an NPM package for string validation and coercion, uses an offensive term as part of its API, namely "blacklist".
I've replaced every usage of that term with "blocklist" in #353 but it's functionally impossible to replace that term when importing that functionality:

[`import blocklist from "validator/lib/blacklist";`](https://github.com/houdiniproject/houdini/blob/35402a0d182b508410793a586c1900cb687edbc1/app/javascript/legacy_react/src/components/create_offsite_payment_pane/CreateOffsitePaymentPane.tsx#L21)

## Possible Fix
We could vendor the function or write our own. We could also see if a package we're already using has an identical function. The two use-cases use legacy tech so we might want to just wait to replace those if we don't use validator in many other places.

faco400:
<h1>Description</h1>
Change the use of a offensive term to a more acceptable name by implementing a function that does the same thing

Closes #354 
Co-authored-by: Rafaelltm <rafaelltm10@hotmail.com>
Co-authored-by: Carlacangussu <karlacangussu@hotmail.com>

<h1>Motivation for implementing it</h1>
The only use of the validator dependency is of the function with the offensive term blacklist. So by implementing another function  with an acceptable term (removeChar) we no longer need this dependency.

<h1>Solution Sumary</h1>

- We created a function named removeChar in app/javascript/legacy_react/src/lib/utils.ts that does the same thing the validator/blacklist does:
```typescript
export function removeChar(i:null|undefined|string, charArray:null|undefined|string) {
  return i.replace(new RegExp('[' + charArray + ']+', 'g'), '')
}
```
- We removed the import of the blacklist in CreateOffsitePaymentPane.tsx and added the new function (removeChar) import: 
`import { castToUndefinedIfBlank, removeChar } from '../../lib/utils';`
- We removed the import of the blacklist in EditPaymentPane.tsx and added the new function (removeChar) import: 
`import { castToUndefinedIfBlank, removeChar } from '../../lib/utils';`
- We changed the use of the blocklist to use the removeChar function in both files to:
`output: (dollarString:string) => parseFloat(removeChar(dollarString, '$,')),`


app/javascript/legacy_react/src/lib/utils.ts
```diff
@@ -15,3 +15,7 @@ export function castToUndefinedIfBlank(i:null|undefined|string) :
     string | undefined {
   return isBlank(i) ? undefined : i;
 }
+
+export function removeChar(i:null|undefined|string, charArray:null|undefined|string) {
+  return i.replace(new RegExp('[' + charArray + ']+', 'g'), '')
+}
```
wwahammy, line 21: This is good work. Would you be able to create some Jest unit tests for this? I think it'd help verify that this works in all of the situations we expect it to.

wwahammy (changes requested):
Nice job! The only request I have is to add some unit tests for the new `removeChar` function.

Really good, and quick, work!

app/javascript/legacy_react/src/lib/utils.ts
faco400: image
@wwahammy Do you know what could be cousing this error? We are trying to run tests for the the removeChar function but for some reason when we run the yarn jest we get this error.

Here is how we are testing:
image

faco400: When we run the test only in this file it passes:
image

app/javascript/legacy_react/src/lib/utils.spec.ts
```diff
@@ -0,0 +1,30 @@
...
+  )
+
+  test("remove occurrences of a char from a string", () => 
+    expect(Utils.removeChar(teststring2,'s')).toBe('thi tring')
+  )
+
+  test("does not remove if not on argument", () => 
+    expect(Utils.removeChar(teststring2,'abcdef')).toBe('this string')
+  )
+
```
wwahammy, line 28: Are there any other conditions you might be forgetting?

If you look at the types of `i` and `charArray` you can receive a string or null or undefined. You should test for those situations and verify the code works as you expect.

Alternatively, should those function arguments accept all of those types?

app/javascript/legacy_react/src/lib/utils.ts
```diff
@@ -15,3 +15,7 @@ export function castToUndefinedIfBlank(i:null|undefined|string) :
     string | undefined {
   return isBlank(i) ? undefined : i;
 }
+
+export function removeChar(i:null|undefined|string, charArray:null|undefined|string) {
```
wwahammy, line 19: Just noticed that there isn't a return type for this function. Make sure to add that.

wwahammy (changes requested):
Nice work on adding the specs. There's a few tweak left.

Also make sure to run `yarn eslint --fix` and fix any errors/warnings that can't be autofixed.

app/javascript/legacy_react/src/lib/utils.spec.ts
faco400: You're right the only type accepted should be string

app/javascript/legacy_react/src/lib/utils.spec.ts
```diff
@@ -0,0 +1,29 @@
+import * as Utils from './utils';
```
wwahammy, line 1: Please add the license header on this file. You can find it at the top of any of the other TS files.

wwahammy (changes requested):
There's one super minor change required, which is to add the license header to the top of the jest spec file. Once that's done and you've pushed it, I'll be happy to merge it!

app/javascript/legacy_react/src/lib/utils.spec.ts
faco400: No problem!

wwahammy (approved):
Looks good! Thanks for all the help!

PR #1247, opened 2022-07-19
Switch to using the latest v1 version of noticeme

wwahammy:
We're now less specific about the version of noticeme we're going to use. We use the latest v1 version of noticeme.


wwahammy:
There's a bug here that I don't quite understand. I need to investigate more.

wwahammy:
I'm going to close this.

PR #1311, opened 2022-08-09
Add eslint-plugin-storybook

wwahammy:
- Add eslint-plugin-storybook
- Add @storybook/testing-library
- Run npx yarn-deduplicate
- Fix the eslint-plugin-storybook errors

Depends on #1310


wwahammy:
We're getting a weird error in #1310 so we'll figure this out later.

PR #1323, opened 2022-08-17
Use nonprofit currency symbol on donation queries

### LINKED ISSUE #1235: [BUG] Rails spec failing based on local currency
When running rails spec in systems that uses another default currency (different from "$"), tests that compare money values happens to fail.

## Description

After setting up the project correctly, when i run the `./bin/rails spec` command, some tests fail, this is clearly due to localization issues (in my case i live in Brazil, and brazilian currency is R$). I don't know if it's some postgres configuration or some computer based configuration. I think the tests should consider any possible currency to run correctly.

Captura de tela de 2022-07-17 18-57-00
Captura de tela de 2022-07-17 18-56-51
Captura de tela de 2022-07-17 18-56-41
Captura de tela de 2022-07-17 18-56-33
Captura de tela de 2022-07-17 18-56-26

## Expected Behavior

* The tests should pass independently of the system currency

## Actual Behavior

* Tests that compares money are failing

## Possible Fix

I don't know if it's a good a ideia, but a possible fix for this issue could be comparing only the value of the money instead of the whole string.

## Steps to Reproduce

**Obs.: The tests only fail in systems that are not configured to use American Dollar as default currency**
1. Set up the project
2. Run `./bin/rails spec`

## Context

This bug happened in the first time I was setting up the project, when the tests failed, my first thought was that something about my set up was wrong and i shoul re-do it. This bug may cause confusion in new contributors to the project from other places in the world.

## Your Environment

* Ruby version: 2.7.6
* Operating System and version: Pop! _OS 20.04

murilogds:
## Description
Uses the nonprofit currency symbol to display the amount of money on donation queries

Closes #1235
Co-authored-by: Giulia Lobo <giulialobo89@gmail.com>

## Motivation to implement it
The amount of money was being displayed based on the currency configured on postgres by using postgres type cast to money:

```ruby
 Qexpr.new.select([
        'donations.created_at',
        '(payments.gross_amount/100.00)::money::text AS amount',
        'COUNT(recurring_donations.id) > 0 AS recurring',
        "STRING_AGG(campaign_gift_options.name, ',') AS campaign_gift_names"
      ]...
```

Because of this, the `rails spec` command ended up failing in systems with different currency from USD, the right behavior in this cases is to use the nonprofit currency symbol instead of postgres type cast.

## Solution summary
Instead of using postgres type cast, we decided to use the `print_currency` existing function from ApplicationHelper, for that we used map based on the query results to change the amount to the currency string displayed.

```ruby
result = QexprQueryChunker.get_chunk_of_query(offset, limit, skip_header) do
      full_search_expr(npo_id, query).select(
        'recurring_donations.created_at',
        'recurring_donations.amount AS amount',
        "concat('Every ', recurring_donations.interval, ' ', recurring_donations.time_unit, '(s)') AS interval",
        '(SUM(paid_charges.amount) / 100.0)::money::text AS total_contributed',
        ...
end
result.map { |r| update_amount_with_currency(r, currency) }
```
update_amount_with_currency function:
```ruby
def self.update_amount_with_currency(query_row, currency)
    if query_row[1] != 'Amount'
      query_row[1] = print_currency(query_row[1], currency, true, true)
    end
    query_row
  end
```

### DIFF
--- a/app/legacy_lib/format/currency.rb
+++ b/app/legacy_lib/format/currency.rb
@@ -23,6 +23,12 @@ module Format
                              .gsub(/^(\d+)\.(\d)$/, '\1.\20') # add a second zero if single decimal (eg. "9.9" -> "9.90")
     end
 
+    # Print money based on informed currency
+    # @param [Integer] cents the amount of money to be printed
+    # @param [String] unit the currency that will be used to print
+    # @param [Boolean] sign whether it should show the sign of the amount
+    # @param [Boolean] use_precision whether it should have two decimal cases
+    # @return [String] the text with the amount of money in the informed currency (Ex.: $10.00)
     def self.print_currency(cents, unit = 'EUR', sign = true, use_precision = false)
       dollars = cents.to_f / 100.0
       dollars = number_to_currency(dollars, unit: unit.to_s, precision: !use_precision && dollars.round == dollars ? 0 : 2)
--- a/app/legacy_lib/format/currency.rb
+++ b/app/legacy_lib/format/currency.rb
@@ -4,6 +4,9 @@
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
 module Format
   module Currency
+    class << self
+      include ActiveSupport::NumberHelper
+    end
     # Converts currency units into subunits.
     # @param [String] units
     # @return [Integer]
@@ -18,5 +21,12 @@ module Format
       (subunits.to_f / 100.0).to_s
                              .gsub(/^(\d+)\.0$/, '\1') # remove trailing zero if no decimals (eg. "1.0" -> "1")
                              .gsub(/^(\d+)\.(\d)$/, '\1.\20') # add a second zero if single decimal (eg. "9.9" -> "9.90")
+    end
+
+    def self.print_currency(cents, unit = 'EUR', sign = true, use_precision = false)
+      dollars = cents.to_f / 100.0
+      dollars = number_to_currency(dollars, unit: unit.to_s, precision: !use_precision && dollars.round == dollars ? 0 : 2)
+      dollars = dollars[1..-1] unless sign
+      dollars
     end
 end; end
--- a/app/legacy_lib/format/currency.rb
+++ b/app/legacy_lib/format/currency.rb
@@ -4,9 +4,6 @@
 # Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE
 module Format
   module Currency
-    class << self
-      include ActiveSupport::NumberHelper
-    end
     # Converts currency units into subunits.
     # @param [String] units
     # @return [Integer]
@@ -31,7 +28,7 @@ module Format
     # @return [String] the text with the amount of money in the informed currency (Ex.: $10.00)
     def self.print_currency(cents, unit = 'EUR', sign = true, use_precision = false)
       dollars = cents.to_f / 100.0
-      dollars = number_to_currency(dollars, unit: unit.to_s, precision: !use_precision && dollars.round == dollars ? 0 : 2)
+      dollars = ActiveSupport::NumberHelper.number_to_currency(dollars, unit: unit.to_s, precision: !use_precision && dollars.round == dollars ? 0 : 2)
       dollars = dollars[1..-1] unless sign
       dollars
     end
--- a/app/legacy_lib/query_donations.rb
+++ b/app/legacy_lib/query_donations.rb
@@ -5,11 +5,19 @@
 
 module QueryDonations
   # Export all donation data for a given campaign
+  class << self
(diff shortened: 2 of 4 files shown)


wwahammy (changes requested):
So I think you're going in the right direction but we're not quite there yet. I think it'd make sense to add a module whose sole method is `print_currency`. Then you can reference that method in ApplicationHelper, QueryDonations and QueryRecurringDonations

I think the most logical place to put this is in the `app/legacy_lib/format/currency.rb` file. From there you should be able to reference it in all three places.

murilogds:
Hi @wwahammy, we made the changes you asked in the code but we don't know how and where should we add the documentation for the print_currency method. Can you help us with that?

wwahammy:
Hi @murilogds, sorry about not getting back to you until now.

You should document the method using the YARD documentation standard. In particularly, the description, the parameters and what is returned. An example of doing so can be found at https://yardoc.org/features.html

murilogds:
> Hi @murilogds, sorry about not getting back to you until now.
> 
> You should document the method using the YARD documentation standard. In particularly, the description, the parameters and what is returned. An example of doing so can be found at https://yardoc.org/features.html

Thanks for the guide @wwahammy I updated the Pull Request with the documentation, can you review it?

app/legacy_lib/query_donations.rb
```diff
@@ -5,11 +5,15 @@
 
 module QueryDonations
   # Export all donation data for a given campaign
+
   def self.campaign_export(campaign_id)
-    Psql.execute_vectors(
+    nonprofit_id = Campaign.find(campaign_id).nonprofit_id
+    currency = Nonprofit.find(nonprofit_id).currency_symbol
```
wwahammy, line 11: I think this would be better as follows:
```suggestion
    currency = Campaign.find(campaign_id).nonprofit.currency_symbol
```

app/legacy_lib/query_donations.rb
```diff
@@ -95,4 +101,11 @@ def self.get_first_payment_for_donation(table_selector = 'donations')
       .order_by('payments.created_at ASC')
       .limit(1)
   end
+
+  def self.update_amount_with_currency(query_row, currency)
+    if query_row[1].to_s.downcase != 'amount'
```
wwahammy, line 106: Please add a comment before this line explaining that you're doing this to skip the header row. It makes sense to do it but I don't think it's obvious just from looking at it.

app/legacy_lib/query_recurring_donations.rb
```diff
@@ -300,4 +303,11 @@ def self.export_for_transfer(nonprofit_id)
     end
     output.to_a
   end
+
+  def self.update_amount_with_currency(query_row, currency)
+    if query_row[1] != 'Amount'
```
wwahammy, line 308: Same comment as above about providing a comment to explain this line.

app/legacy_lib/format/currency.rb
```diff
@@ -19,4 +22,17 @@ def self.cents_to_dollars(subunits)
...
+
+    # Print money based on informed currency
+    # @param [Integer] cents the amount of money to be printed
+    # @param [String] unit the currency that will be used to print
+    # @param [Boolean] sign whether it should show the sign of the amount
+    # @param [Boolean] use_precision whether it should have two decimal cases
+    # @return [String] the text with the amount of money in the informed currency (Ex.: $10.00)
+    def self.print_currency(cents, unit = 'EUR', sign = true, use_precision = false)
+      dollars = cents.to_f / 100.0
+      dollars = number_to_currency(dollars, unit: unit.to_s, precision: !use_precision && dollars.round == dollars ? 0 : 2)
```
wwahammy, line 34: Does this work with you use `ActiveSuppor::NumberHelper.number_to_currency` instead? If so, I think it'd make more sense to use than and not do the include of `ActiveSupport::NumberHelper`. We're only using that single method from `ActiveSupport::NumberHelper` and only using it in one place.

app/legacy_lib/query_donations.rb
```diff
@@ -5,11 +5,15 @@
 
 module QueryDonations
   # Export all donation data for a given campaign
+
```
wwahammy, line 8: Super minor tweak but could you remove this new line break?

wwahammy (changes requested):
Very good work! There's a few minor tweaks needed but otherwise, it's almost ready to merge.

wwahammy:
Hi @murilogds @Giuulob89, just checking in. Is this ready for review?

murilogds:
Sorry for the delay @wwahammy I forgot to ask your review

wwahammy (approved):
Looks good to me!

wwahammy:
Thanks for all the help  @murilogds @Giuulob89!

PR #1345, opened 2022-08-26
Add a dependency review to prevent new vulnerabilities from being added

wwahammy:


wwahammy:
This doesn't really do much of anything so let's scrap it.

PR #1365, opened 2022-09-13
Upgrade msw

wwahammy:
msw has been upgraded a few times since it was installed so let's update it!


wwahammy:
This breaks all of the tests for some reason so we'll close for now.

PR #1368, opened 2022-09-14
Alphabetize the properties in jest.config.js

wwahammy:
Makes it easier to notice if anything has changed in the Jest config.

Also removed some no longer used config for ts-jest.


wwahammy:
This keeps failing and I don't know why so I'm just going to close and be done with it.

PR #1391, opened 2022-10-24
Update webpack-dev-server-to-4

wwahammy:


wwahammy:
I don't think this is the right version to use so let's close and deal with later.

Issue #3, opened 2018-03-25
Take saved info out of public and make that folder generated data only


ericschultz:
In the repo, a few files are saved in `public`. Since this folder should be all generated info, this is a problem. These files should be moved into other folders as appropriate and then copied in via Webpack.

Additionally, after the files are moved, the `public` directory should be fully gitignore'd.

fellipepcs:
Hello, DafneMoretti and I will be working on this issue to organize the public folder and gitnote.

PR #1416, opened 2022-11-28
Resolve public folder fix

DafneM:
* Moved not generated files into other folders as appropriate
* Gitignore'd public directory

Closes issue: https://github.com/houdiniproject/houdini/issues/3

Co-Authored-by: fellipepcs <fellipy.02@hotmail.com>

@wwahammy and @clarissalimab for some reason, I couldn't add you as reviewers. Can you review this for us?


wwahammy:
Hi @DafneM and @fellipepcs thank you for your contribution and welcome to Houdini! This a great start to jump in.

I think there are some potential pitfalls to be careful with here and we need to verify that those things are addressed. Here's a few issues that we need to be sure of:

* {404, 422, 500}.html are the default fallbacks for when those errors are returned when visiting a page. Does that continue to be the case after the change?

* favicon.ico is the favicon that for use by browsers. Browsers can get favicons a couple different ways:
** go to <domain>/favicon.ico
** from the favicon link in the head of a particular webpage. For example: the link might look like `<link rel="icon" href="favicon.ico" type="image/x-icon" />`. In Rails, this is added to a particular template using the `favicon_link_tag` helper.
 
 Are we sure the favicon still works?

* Are the various moved logos reference to in any place in the code that needs to be adjusted? 

* Are the images originally in  `public/images/fallback/` or `public/fallback/`

I think a good first part of this would be to address the error pages first as a separate PR. The Rails 6.1 guides should have information on how to set error pages. As you work on that, make sure you understand what the current behavior is, what the goals of the change would be (for example, why do we want to move these things? Moving the files is work so there must be a reason). If possible, I would encourage having unit tests to verify the current behavior and then, after the change, verify that the behavior continues to work. You could probably do this via a request spec.

I hope that's all helpful. Once again, thanks so much for the contributions so far; I'm excited to see what you're able to come up with!

fellipepcs:
> 

Goodnight!  @wwahammy 

I would like you to better specify this part of the favicon, and what is the next step after that.

PR #1439, opened 2022-12-14
Resolve errors page fix

fellipepcs:
# Descrição

- Adicionado páginas de erro que podem ser personalizadas dos status 404,422 e 500 do Rails.
- Criado Classe ErrorsController para utilização em diversos locais onde deseja implementar as paginas de erros, chamando apenas o método. 

Complementa o problema: https://github.com/houdiniproject/houdini/issues/3

Realizado por @DafneM e @fellipepcs


config/routes.rb
```diff
@@ -290,6 +290,13 @@
   get '/css/donate-button.css' => 'widget#v1_css'
   get '/css/donate-button.v2.css' => 'widget#v2_css'
 
+  # Routes
```
wwahammy, line 293: I think this would be better if it was "Error Routes"

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ErrorsController < ApplicationController
```
wwahammy, line 1: Since these are new files, you'll need to add the following to indicate you're licensing your code under our license and that you want to use frozen string literals
```ruby
# frozen_string_literal: true

```

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def internal_server
+        render status: 500
+    end
+
+    def unprocessable
+        render status: 422
+    end
+
+    def unaccetable
```
wwahammy, line 14: Wanted to mention this is a typo and should be unacceptable.

app/controllers/errors_controller.rb
```diff
@@ -0,0 +1,17 @@
+class ErrorsController < ApplicationController
+    def not_found
+        render status: 404
+    end
+
+    def internal_server
```
wwahammy, line 6: Please change this to `internal_server_error`

app/views/errors/internal_server.html.erb
```diff
@@ -0,0 +1,3 @@
+<%# app/views/errors/internal_server.html.erb %>
```
wwahammy, line 1: Same thing regarding the licensing here. Please replace these lines in the top of the new erb files with:

```ruby
<%- # License: AGPL-3.0-or-later WITH WTO-AP-3.0-or-later
# Full license explanation at https://github.com/houdiniproject/houdini/blob/main/LICENSE -%>
```

app/views/errors/not_found.html.erb
```diff
@@ -0,0 +1,3 @@
+<%# app/views/errors/not_found.html.erb %>
+
+This page does not exit. Please go back
```
wwahammy, line 3: Please use `Oops... couldn't find that page` as in the original 404.html

config/routes.rb
```diff
@@ -290,6 +290,13 @@
...
+
+  get '/404', to: 'errors#not_found'
```
wwahammy, line 295: Minor change: could these error routes be in numerical order? It'd be easier for maintenance.

app/controllers/errors_controller.rb
wwahammy, line 14: Also, there's no route or view for this method. So it should be deleted or the appropriate route and view should be added.

app/controllers/errors_controller.rb
wwahammy, line 1: Normally, you always want to inherit from ApplicationController but this is a special case: let's inherit from `ActionController::Base`. The reason is that if there's a bug actually in ApplicationController, it might not be possible to load error pages if we inherit from it.

public/404.html
```diff
@@ -1,49 +0,0 @@
-<!--  License: CC0-1.0 -->
```
wwahammy, line 1: I'd like the new code error pages styled somewhat like these original pages. Here's what it needs:

* the title and appropriate error code
* the body looking relatively similar but:
  *  use the appropriate erb and rails view functionality using [View Helpers](https://guides.rubyonrails.org/v6.1/action_view_helpers.html) and other necessary parts of ActionView.
  * use  `image_tag` for [including the image](https://guides.rubyonrails.org/v6.1/action_view_helpers.html#assettaghelper). This means the SVG will have to be in the appropriate place in the `app/assets` directory (Also, make sure to add an `alt` description to that image; it should have been there in the first place.
  
  *move to the head of the document like it's supposed to be.
  
 * create and use a layout just for ErrorController. The goal is to be able to reuse everything between all of the error pages except the error messages unique to each error page.

wwahammy (changes requested):
@DafneM and @fellipepcs thank you so much for the contribution. I particularly appreciate the initiative you took to realize that we should use `exceptions_app`. This is a really great start.

I've made a number of comments on necessary improvements before merging; most of them are minor. I look forward to your future changes. Please re-request my review when you're ready.

Thanks so much again.

fellipepcs:
Made changes to requests requested above.

DafneM:
Hey @wwahammy is there anything we can do here?

PR #1441, opened 2022-12-17
Updating readme with rbenv

joaobisi:


joaobisi:
@wwahammy We just realized that we forgot to include the ruby ​​installation step. We made the change in this pull request. Sorry for the mistake.

wwahammy:
@joaobisi thanks for this. Good catch!

PR #1460, opened 2023-01-12
the method refund notification sends emails only if the array named emails has any occurrences

joaobisi:
…mails has any occurrences

Co-authored-by: kayrocesar <kayro.cesar.kc@gmail.com>
Co-authored-by: victordscabral <victorcabral2501@gmail.com>


joaobisi:
@wwahammy I'm not shure if that works. Tell us if it has any problem.

wwahammy:
It probably does but it would be good to add a test to verify it. Unfortunately, we don't have a mailer spec for NonprofitMailer to easily test this. I will merge but will also add an issue for adding a spec for the mailer.

PR #1472, opened 2023-01-17
button spacing fix

kayrocesar:
Co-Authored-By: Victor Cabral <87657291+victordscabral@users.noreply.github.com>
Co-Authored-By: J <111386960+joaobisi@users.noreply.github.com>

before:
telaantes

after:
depois

### DIFF
--- a/app/javascript/legacy/nonprofits/donate/info-step.js
+++ b/app/javascript/legacy/nonprofits/donate/info-step.js
@@ -106,9 +106,9 @@ function view(state) {
   , customFields(state.params$().custom_fields)
   , dedicationLink(state)
   , app.nonprofit.no_anon ? '' : anonField(state)
-  , h('fieldset.u-inlineBlock.u-marginTop--10.u-marginLeft--65', paymentMethodButtons(["card", "sepa"], state))
+  , h('fieldset.u-inlineBlock.u-marginTop--10', paymentMethodButtons(["card", "sepa"], state))
   ])
-  return h('div.wizard-step.info-step.u-padding--10', [
+  return h('div.wizard-step.info-step.u-padding--10.centered', [
     form
   , h('div', {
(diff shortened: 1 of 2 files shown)


wwahammy:
Thanks for the submission and welcome to Houdini. I only have one request: could you center the buttons? So definitely split them up by the same distance but put them in the center of that line.

Otherwise it looks good!

kayrocesar:
New fix:

info_buttons

app/javascript/legacy/nonprofits/donate/info-step.js
```diff
@@ -106,7 +106,7 @@ function view(state) {
   , customFields(state.params$().custom_fields)
   , dedicationLink(state)
   , app.nonprofit.no_anon ? '' : anonField(state)
-  , h('fieldset.u-inlineBlock.u-marginTop--10', paymentMethodButtons(["card", "sepa"], state))
+  , h('fieldset.u-inlineBlock.u-marginTop--10.u-marginLeft--65', paymentMethodButtons(["card", "sepa"], state))
```
wwahammy, line 109: Have you tried using something that sets `text-align: center` on this element? I think that would be the most ideal.

kayrocesar: We tried with solutions like that, but some problems occurred. Now that we have a better understanding of the structure of the project we will implement a solution of the type. Thanks for the feedback!

wwahammy: @kayrocesar when you say problems occurred, could you describe them in more detail? It's good to have it documented and also so I can help you and you can help me.

kayrocesar: Hello @wwahammy, sorry for the delay in answering. 
We tried to use text-align but it didn't work before, because we were doing it in the wrong place. But we were able to align the buttons using it now, if you could take a look. Thanks a lot.

kayrocesar:
Comparison:

buttons_before
buttons_center

PR #1474, opened 2023-01-17
[WIP] Feat/add ansible 

Maiconrq:


wwahammy:
Hi @Maiconrq and @wellpriz, thanks for your interest in Houdini. While I appreciate the interest, I don't think we really want an Ansible set up for Houdini. It's not necessarily a bad thing to have but someone needs to maintain it over time. Unless someone is willing to do so, I can't really merge it.

That said, it wouldn't be a bad thing to document how to do this on the [Houdini Project discussions ](https://github.com/houdiniproject/houdini/discussions). That way the information is available still for those who work on it.

PR #1488, opened 2023-01-31
Added changes to dependabot for creating labels

fellipepcs:
# Description

Added labels in dependabot file, and created labeler file for creating labels in actions.

Co-authored-by: DafneM <moretti.dafne2@gmail.com>


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,16 @@
...
+  - pull_request
+
+permissions: write-all
+
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+
+      - uses: srvaroa/labeler@v0.8
```
wwahammy, line 12: Could you describe why you picked this specific action here? Github builds their own (`actions/labeler`). It would be helpful if you explained the pros and cons of that versus this one in making your decision.

config/routes.rb
```diff
@@ -1,372 +0,0 @@
-# frozen_string_literal: true
```
wwahammy, line 1: If you removed all of these files, the entire app would not work, was there a reason they were removed?

wwahammy (changes requested):
Thanks for the PR @fellipepcs. There's a two big things that need to be addressed here. First, we need all of those files you deleted, those are important.

Second, I'd like a better sense of why you picked the particular action for labeling.

PR #1521, opened 2023-04-04
Use unlogged Postgres tables in testing

wwahammy:


wwahammy:
This doesn't seem to have a significant effect on our test performance so I'm closing.

PR #1543, opened 2023-05-03
Create a not_change rspec matcher for compound expectations

wwahammy:
This adds the ability to check expectations like so:

```ruby
expect { fire_error_instead_of_creating_user}.to raise_error(ExpectedError).and not_change { User.count }
```


wwahammy:
Oops, wrong repo

Issue #1427, opened 2022-12-06
[FEATURE] Use noticeme based on a cron schedule


wwahammy:
ClearlyDefined is a great tool but it is a MASSIVE headache to run on every PR. This is due to two problems:

1.  ClearlyDefined is often overloaded
2. Notices will change in between PRs so people's PRs break for now reason

This would move noticeme verification and updating to running on a regular schedule. For closing, this change will:

- [ ] Remove the noticeme run from inside the PR tasks
- [ ] Create a new workflow for running noticeme that checks if there are changes and, if so, creates a PR and notifies the team

One advantage of doing this is that we can more easily use [Dependabot batcher](https://github.com/Legal-and-General/dependabot-batcher) which would save us some work.

victorear05:
I'll work on it!

wwahammy:
@victorear05 Thanks! Please use the changes I just made in react-i18n-currency-input as a guide. The noticeme workflow is viewable at https://github.com/houdiniproject/react-i18n-currency-input/blob/main/.github/workflows/noticeme.yml

PR #1555, opened 2023-05-22
Use noticeme based on a cron schedule (#1427)

victorear05:
I've changed the "noticeme" for using like was done in:
https://github.com/houdiniproject/react-i18n-currency-input/blob/main/.github/workflows/noticeme.yml


victorear05:
The solution didn't work

wwahammy:
I just looked at this and honestly, you were pretty close.

victorear05:
@wwahammy, Vitor, can you tell me how I can test a solution? It's because in my fork, even the job that already ran today, is failing

wwahammy:
@victorear05 I'll try to explain what I think the solution in a review

.github/workflows/noticeme.yml
```diff
@@ -0,0 +1,75 @@
...
+    - name: configure git
+      run: |
+        git config user.name "notice-update@houdiniproject.org"
+        git config user.email "notice-update@houdiniproject.org"
+    - name: commit NOTICE-ruby if needed
+      run: |
+        if ! git diff --exit-code NOTICE-ruby; then
+          git add NOTICE-ruby
+          git commit -m 'Update NOTICE for ${{ env.date }}'
+          git checkout -b notice-update-$date
```
wwahammy, line 72: I think it would help if the local and remote branch names matched

```suggestion
          git checkout -b notice-update-ruby-$date
```

wwahammy (commented):
Give a few of these changes a try an dsee if it helps.

PR #1596, opened 2023-07-10
Remove duplication of prerequisites and testing in docs

BenSturmfels:


BenSturmfels:
I've assumed `bin/rails spec` is preferred over `bundle exec rspec` as that's what's in `README.md`. Is that right?

BenSturmfels:
Unrelated point - I think it would make sense to move all the detailed local development install and configure docs to `docs/getting_started.md` and leave `README.rst` as an overview of the project, how to get involved etc. What do you think?

wwahammy:
> I've assumed `bin/rails spec` is preferred over `bundle exec rspec` as that's what's in `README.md`. Is that right?

So, it depends :)

Quick point: `bin/rails spec` and `bundle exec rails spec` are practically identical, the `bin` version actually calls the `bundle `exec` version. Same with `bin/rspec` and `bundle exec rspec`.

If you're running the test suite, `bin/rspec` and `bin/rails spec` should be identical I think. The only exception is if the test database isn't setup already, `bin/rails spec` will do that on first run. On the other hand `bin/rspec` provides all of the features of rspec. The major one I like is `--only-failures` which will run only failed tests from the last complete test run. 

`--next-failure` is similar and will run failed tests from the last complete test run and stop at the first failure. Additionally, it will skip any tests which you've gotten working on a previous `--next-failure` run. It's really helpful for going through broken tests in turn.

I hope that helps!

README.md
```diff
@@ -35,11 +35,10 @@ All new backend code and React components should be well tested.
 
 ## Prerequisites
 
-- Node 14
+- Node.js 14 (we require 14 because we want the full internationalization built-in)
 - Yarn
-- PostgreSQL 10  or 12
+- PostgreSQL 12 (10 probably works)
 - Ruby 2.7
-- Ubuntu 18.04, 20.04, 22.04 or equivalent
```
wwahammy, line 42: I'm a little hesitant to remove this entirely. The point is we're not supporting non-Ubuntu/Debian versions. Maybe a different phrasing can be used though.

wwahammy (changes requested):
I have one tweak I'd like made but otherwise this is very helpful!

wwahammy:
> Unrelated point - I think it would make sense to move all the detailed local development install and configure docs to `docs/getting_started.md` and leave `README.rst` as an overview of the project, how to get involved etc. What do you think?

I'm open to the idea. I tend to prefer it in the main README but I'm fine either way.

README.md
BenSturmfels: No problem - Ubuntu reinstated it latest commit.

BenSturmfels:
> Quick point: `bin/rails spec` and `bundle exec rails spec` are practically identical, the `bin` version actually calls the `bundle `exec`version. Same with`bin/rspec`and`bundle exec rspec`.
> 
> If you're running the test suite, `bin/rspec` and `bin/rails spec` should be identical I think. The only exception is if the test database isn't setup already, `bin/rails spec` will do that on first run. On the other hand `bin/rspec` provides all of the features of rspec. The major one I like is `--only-failures` which will run only failed tests from the last complete test run.
> 
> `--next-failure` is similar and will run failed tests from the last complete test run and stop at the first failure. Additionally, it will skip any tests which you've gotten working on a previous `--next-failure` run. It's really helpful for going through broken tests in turn.

Thanks, I've mentioned this in a subsequent commit.

BenSturmfels:
> > Unrelated point - I think it would make sense to move all the detailed local development install and configure docs to `docs/getting_started.md` and leave `README.rst` as an overview of the project, how to get involved etc. What do you think?
> 
> I'm open to the idea. I tend to prefer it in the main README but I'm fine either way.

No worries - I'll leave as-is for now.

wwahammy (approved):
Thanks @BenSturmfels!

PR #1597, opened 2023-07-10
Document how to clear the built CSS/JS assets

BenSturmfels:


docs/getting_started.md
```diff
@@ -80,6 +80,9 @@ reloading, run `bin/webpack-dev-server` in a second console. Now after making a
 change to the JavaScript code you should see Webpack immediately rebuild the
 assets and perform a full page reload.
 
+When in doubt, you can manually clear the generated CSS and JavaScript files,
+run `rm -rf public/assets public/packs`.
```
wwahammy, line 84: There's actually a [built in command ](https://guides.rubyonrails.org/v6.0/command_line.html#rails-assets) for that: `bin/rails assets:clobber`. That's the appropriate way to handle that.

wwahammy (changes requested):
Thanks @BenSturmfels. There's actually a built-in command for this as described.

docs/getting_started.md
BenSturmfels: Thanks Eric, I've made that change.

wwahammy (approved):
Thanks @BenSturmfels!

Issue #540, opened 2021-03-12
[BUG] ERROR:  relation "active_storage_blobs" does not exist


kenspy:
Came across this issue while installing the v2.

## Error
```
-- column_exists?(:active_storage_blobs, :service_name) 
rails aborted! 
StandardError: An error has occurred, this and all later migrations canceled: 
 
PG::UndefinedTable: ERROR:  relation "active_storage_blobs" does not exist 
LINE 8:  WHERE a.attrelid = '"active_storage_blobs"'::regclass 
 
/home/xxxxx/www/db/migrate/20210105192021_add_service_name_to_active_storage_blobs.active_storage.rb:4:in `up' 
 
Caused by: 
ActiveRecord::StatementInvalid: PG::UndefinedTable: ERROR:  relation "active_storage_blobs" does not exist 
LINE 8:  WHERE a.attrelid = '"active_storage_blobs"'::regclass 
```

## Fixed
`rails active_storage:install`

## Your Environment

* Ruby version: 2.7.2
* Operating System and version: Ubuntu

kenspy:
Hmm, despite active_storage successful installation, I am still unable to migrate, the error above still persist. Do I need to configure storage somewhere?

wwahammy:
@kenspy No, you shouldn't need to set anything up for storage, it just uses a directory `/storage` under the root.

Let me try to set this up in a separate VM and build from scratch and see if I can replicate this.

kenspy:
@wwahammy Any luck reproducing the error?

wwahammy:
@kenspy, I looked at it on Friday and couldn't replicate the error. I'm not quite sure what's going on.

I think the best course of action would be to clear the database and start over. To do so run the following:

```ruby
bin/rails db:drop db:create # deletes the db and recreates it
bin/setup # loads the db
```

Let me know if that doesn't work.

BenSturmfels:
I'm just trying to revive my Houdini development environment and hit this errer `PG::UndefinedTable: ERROR:  relation "active_storage_blobs" does not exist`. This occurs for me with a fresh database when running `bin/setup` - console output attached.

[setup-log.txt](https://github.com/houdiniproject/houdini/files/11540336/setup-log.txt)

wwahammy:
@BenSturmfels thanks for that, I was able to recreate it!

The problem is due to a quirk about how `bin/rails db:prepare` works when called in `bin/setup`.  `db:prepare` does one of two things:

[...]

PR #1598, opened 2023-07-10
Document workaround for `relation "active_storage_blobs" does not exist`

BenSturmfels:
See #540


wwahammy (approved):
Thanks @BenSturmfels, this is very helpful!

PR #1631, opened 2023-08-26
Fix bug on sending emails

wwahammy:


wwahammy:
@BenSturmfels this should fix the annoying problem you noticed later in #1620

BenSturmfels:
> @BenSturmfels this should fix the annoying problem you noticed later in #1620

Thanks @wwahammy, that's working well now!

PR #1673, opened 2023-11-17
Initial Step Manager work

wwahammy:
**NOTE: DO NOT discuss internal CommitChange information in your PR; this PR will be public.
Link back to the issue in the Tix repo when you need to do that.**


wwahammy:
This was in the wrong repo

PR #1817, opened 2025-02-17
Switch to standard

wwahammy:
- **Switch to standard**
- **Add .standard.yml file**
- **Run bin/standardrb --fix**


wwahammy:
Handled by #1831

PR #1823, opened 2025-03-13
Cleanup Routes

caseyhelbling:
Quick PR to clean up routes file
* remove random unnecessary whitespace
* tabs to spaces
* use member and collection blocks instead of duplicating passed args
* use shorter symbol array syntax


caseyhelbling:
Uggg... Wrong repo.

